Issue #2093, opened 2025-10-23
AIIDA: Add CIM RTD 1.12


tformatix:
[RealTimeData Document_v1.06..zip](https://github.com/user-attachments/files/23095870/RealTimeData.Document_v1.06.zip)

PR #2112, opened 2025-10-31
GH-2093 Add CIM RTD 1.12

stefanpenzinger:
- Added CIM RTD v1.12
- Move AIIDA asset and schema to API module
- Schema Formatters share a common interface and base class
   - Spring registers Schema Formatters automatically
   - The CIM Formatter uses a strategy pattern to add future CIM versions without complications
   - Remove the need for Util classes
- Use MQTT connected to subscribe to all existing permissions on startup (prevent race condition)

### DIFF
--- a/aiida/src/main/resources/db/aiida/migration/V7_4__add_version_to_smart_meter_p1_cim_schema.sql.license
+++ b/aiida/src/main/resources/db/aiida/migration/V7_4__add_version_to_smart_meter_p1_cim_schema.sql.license
@@ -1,2 +1,0 @@
-SPDX-FileCopyrightText: 2026 The EDDIE Developers <eddie.developers@fh-hagenberg.at>
-SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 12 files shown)


aiida/src/main/resources/db/aiida/migration/V7_4__add_version_to_smart_meter_p1_cim_schema.sql.license
fweingartshofer: You can include the copyright notices directly in the migrations.

data-needs/src/main/resources/db/migration/data-needs/V2_0__add_version_to_smart_meter_p1_cim_schema.sql.license
fweingartshofer: Same as before, you can include this directly in the migration.
The reason why I did it separately, is that I didn't want to change existing migrations, which would have broken flyway migrations.

fweingartshofer (commented):
That is quiet a large PR, next time I think you split all the different outbound connectors into smaller PRs to keep the mental load lower.

There are only some small issues, mostly one type, then this is good to be merged.

Issue #2144, opened 2025-11-11
AIIDA - FR data connector - map additional fields to OBIS codes


rustyb:
Would it be possible to update the french connector in AIIDA to map more of the fields to OBIS codes? 

I believe these mappings are required for the conversion to the SMART-METER-P1 message format

I am currently recieving the following message with the following dataneed configuraiton for a FR teleinfo device

data need

```json
{
  "type": "outbound-aiida",
  "id": "e3d8c4bd-e83c-434f-ad05-a8a9d857d9c4",
  "name": "D4G Realtime Data New Aiida Cim",
  "description": "Send your realtime meter power and energy in and out to Digital4Grids platform.",
  "purpose": "Your data will be used for preparing baselines and proposing flex opportunities for your devices.",
  "policyLink": "https://www.example.com/",
  "createdAt": "2025-09-07T21:30:35.490298Z",
  "enabled": true,
  "duration": {
    "type": "relativeDuration",
    "start": "P0D",
    "end": "P3M"
  },
  "transmissionSchedule": "*/2 * * * * *",
  "schemas": ["SMART-METER-P1-CIM"],
  "asset": "CONNECTION-AGREEMENT-POINT",
  "dataTags": []
}

```

MQTT message arriving at Eddie EMQX which contains only these values:

- TotalActiveEnergyProduced_export_kWh
- InstantaneousVoltage_V_on_phaseL1
- InstantaneousCurrent_A_on_phaseL1

```json
{
  "messageDocumentHeader.creationDateTime": "2025-11-11T03:34:52Z",
  "messageDocumentHeader.metaInformation.connectionId": "colin_testing",
  "messageDocumentHeader.metaInformation.dataNeedId": "e3d8c4bd-e83c-434f-ad05-a8a9d857d9c4",
  "messageDocumentHeader.metaInformation.documentType": "near-real-time-market-document",
  "messageDocumentHeader.metaInformation.permissionId": "689c2846-36c7-4267-857a-d2a303169ac5",
  "messageDocumentHeader.metaInformation.finalCustomerId": "37eac929-5ca5-4318-9d75-a17db7eb8455",
  "messageDocumentHeader.metaInformation.asset": "CONNECTION-AGREEMENT-POINT",
  "messageDocumentHeader.metaInformation.dataSourceId": "d590235e-d838-4346-993c-7bb3720d1e39",
  "messageDocumentHeader.metaInformation.region.connector": "aiida",
[...]

Issue #2106, opened 2025-10-28
AIIDA: Add asset information to data source


tformatix:
## Short Description

Currently, data sources can only store the asset type but not other asset information like the operatorId and the meterId.
To ensure proper identification and traceability, data sources must support adding the operator ID and the physical meter ID (**OPTIONAL**).

This feature is essential for **FCA** use cases, as the CSU must know which data belongs to which physical meter.
Later, this information will enable the **FCA Cockpit** to visualize FCAs per **substation** and **transformer station**, which is crucial for grid stability.

A single smart meter data source can:

- [ ] One physical meter ID
- [ ] One operator ID

This information must be used in the CIM RTD schema.

## Blocked

- [x] #2306

## Acceptance Criteria
- [ ] One physical meter ID can be added to a data source.
- [ ] One operator ID can be added to a data source.
- [ ] The fields are optional.
- [ ] IDs are included in the outbound data schema.

## Design/UI/UX Requirements

- [ ]  When creating or editing a data source, the UI must provide a clear and user-friendly way to add a physical meter ID and operator ID.
- [ ] The data source list view must display associated physical meter ID and operator ID.

PR #2275, opened 2025-12-19
GH-2144 Map additional fields to OBIS codes for FR data connector

Flo-Auzi:
- Map CCASN and CCAIN codes to respective OBIS codes
- Pass PRM instead of DataSourceId as mRID
- Update StandardMode test for MicroTeleinfoV3Adapter


tformatix (changes requested):
Wait until #2306 is exactly clarified and probably further extend it with #2106. We might put it anyways into the new CIM version.

PR #2295, opened 2026-01-02
NOISSUE Use Wiener Netze for Eda E2E tests as they are now usually faster to respond

re1:


re1:
Closing due to no significant improvement in reliability.

Issue #2104, opened 2025-10-28
AIIDA: Add option to state MULTIPLE dataNeedIds for a single Connect Button


tformatix:
## Short Description
Currently, the permission facade supports only one data need.
For flexible connection agreements (FCAs), the eligible party needs to request both an inbound and outbound AIIDA data need. Instead of requesting both permissions separately, it should be possible to pass both data needs to the permission facade, leading to **multiple permission requests**.

A list of permission requests can be accepted/rejected simultaneously. 
Therefore, the QR code needs to be extended with a list of permissions.

Data need to permission (request) is a one-to-one relationship.

### QR code

```json
{
  "eddieId": "56faf485-ed0a-4066-8cce-dc227c1350a9",
  "permissions": [
    {
      "permissionId": "56faf485-ed0a-4066-8cce-dc227c1350a8",
      "serviceName": "My Data Need"
    },
    {
      "permissionId": "56faf485-ed0a-4066-8cce-dc227c1350a7",
      "serviceName": "My other Data Need"
    }
  ],
  "handshakeUrlTemplate": "https://framework.eddie.energy/region-connectors/aiida/permission-request/handshakeUrlTemplate",
  "bearerToken": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJleHAiOjE3NjUzNjkyODMsImlhdCI6MTc2NDc2NDQ4MywicGVybWlzc2lvbnMiOnsiYWlpZGEiOlsiMzQ4NmUxNTktODExNy00ZDQ4LTk0YzItNTU0NmEyZjFjYzY2Il19fQ.JIWhKa2ZinPyYqx_A4rxMmr2sg2wWWdTBr2z220u2RI"
}
```

## Acceptance Criteria
- [x]  Multiple permission requests can be created simultaneously.
- [x]  Accepting the list of permission requests accepts all at once.
- [x]  Each accepted data need results in a separate permission.

## Design/UI/UX Requirements

- [x] Display multiple data needs in the EDDIE button.
- [x] Display multiple data needs in the AIIDA UI.

PR #2320, opened 2026-01-22
GH-2104 Add option to create multiple permission requests with a single connect button in AIIDA

tformatix:


re1 (approved):
Tested locally and worked fine.

Issue #2305, opened 2026-01-09
Admin Console: Adjust existing features to match the Figma design


re1:
## Short Description
The admin console received a [Figma design](https://www.figma.com/design/KKHDSHrFkyCD5UAV4DSKTL/Admin-Console?node-id=0-1&t=JlUYokq5Xk4iSfZy-1) which provides a visual refresh, refines existing features, and also defines planned features not yet implemented.

This task is only concerned with adjusting existing features to match the new design. Features present in the design that are not yet implemented should be implemented with their respective issues. Additional issues are to be created when needed.

The current implementation makes use of PrimeVue components. Either these components are to be themed to match the design, or we implement our own components to replace them.

## Acceptance Criteria
- [ ] Design tokens are configured for light and dark modes
- [ ] Existing features are adapted to match the Figma design

## Design/UI/UX Requirements

<img width="9920" height="3507" alt="Image" src="https://github.com/user-attachments/assets/b7f67d92-51ff-4486-91d6-37451a217fee" />

<img width="9155" height="3507" alt="Image" src="https://github.com/user-attachments/assets/a39389c1-0564-4f64-86e7-2edd9ea2a1d5" />

PR #2348, opened 2026-02-04
GH-2305 Adjust region connectors view to match Figma

re1:
The main goal of this PR is to adjust the region connectors view to match Figma. However, I included some additional related changes.

- Fixed region connector health status.
- Added a helper for parsing the ISO duration coming from the server (including negative durations, which are not supported by JS' Temporal API). We already used the tinyduration package on the EDDIE Button for a similar use-case.
- Adjusted font-sizes and panel borders, which also affects the dashboard.
- Made the sidebar sticky floating instead of scrolling the main container. Helps me with doing full-page screenshots. Will revert if something breaks.

The PR does not include the configuration table shown in Figma. I am still figuring out if we want to use the actuator API here and how to handle sensitive information.

Dark mode button does not have the gradient yet. Doing buttons collectively later.

`RegionConnectorsView.vue` is large and quite convoluted. Did not want to extract components early to maintain locality of behavior, and did not bother to find a way to make it nicer to be honest. Feel free to bring up maintainability issues anyway.

Frontend code mainly concerns @GoodVibezOnly. @fweingartshofer, feel free to check files changed for fixes.

---

[Figma link](https://www.figma.com/design/qPgGqrrmS7ECiqWQz25r9T/Admin-Console?node-id=9819-35235&m=dev) for reference.

image

---

**Screenshots**

image

image

### DIFF
--- a/.idea/copyright/profiles_settings.xml
+++ b/.idea/copyright/profiles_settings.xml
@@ -3,6 +3,10 @@
     <module2copyright>
       <element module="All" copyright="Apache 2.0 SPDX" />
     </module2copyright>
+    <LanguageOptions name="JAVA">
+      <option name="fileTypeOverride" value="3" />
+      <option name="block" value="false" />
+    </LanguageOptions>
     <LanguageOptions name="JavaScript">
       <option name="fileTypeOverride" value="3" />
       <option name="block" value="false" />
--- a/core/src/main/java/energy/eddie/spring/regionconnector/extensions/RegionConnectorSupportedFeatureExtension.java
+++ b/core/src/main/java/energy/eddie/spring/regionconnector/extensions/RegionConnectorSupportedFeatureExtension.java
@@ -1,7 +1,5 @@
-/*
- * SPDX-FileCopyrightText: 2024-2026 The EDDIE Developers <eddie.developers@fh-hagenberg.at>
- * SPDX-License-Identifier: Apache-2.0
- */
+// SPDX-FileCopyrightText: 2024-2026 The EDDIE Developers <eddie.developers@fh-hagenberg.at>
+// SPDX-License-Identifier: Apache-2.0
 
 package energy.eddie.spring.regionconnector.extensions;
 
--- a/core/src/test/java/energy/eddie/spring/regionconnector/extensions/RegionConnectorSupportedFeatureExtensionTest.java
+++ b/core/src/test/java/energy/eddie/spring/regionconnector/extensions/RegionConnectorSupportedFeatureExtensionTest.java
@@ -1,7 +1,5 @@
-/*
- * SPDX-FileCopyrightText: 2024-2026 The EDDIE Developers <eddie.developers@fh-hagenberg.at>
- * SPDX-License-Identifier: Apache-2.0
- */
+// SPDX-FileCopyrightText: 2024-2026 The EDDIE Developers <eddie.developers@fh-hagenberg.at>
+// SPDX-License-Identifier: Apache-2.0
 
 package energy.eddie.spring.regionconnector.extensions;
 
--- a/e2e-tests/src/test/java/energy/eddie/e2etests/adminconsole/AdminConsoleTest.java
+++ b/e2e-tests/src/test/java/energy/eddie/e2etests/adminconsole/AdminConsoleTest.java
@@ -1,7 +1,5 @@
-/*
- * SPDX-FileCopyrightText: 2025-2026 The EDDIE Developers <eddie.developers@fh-hagenberg.at>
- * SPDX-License-Identifier: Apache-2.0
- */
+// SPDX-FileCopyrightText: 2025-2026 The EDDIE Developers <eddie.developers@fh-hagenberg.at>
+// SPDX-License-Identifier: Apache-2.0
 
 package energy.eddie.e2etests.adminconsole;
 


admin-console/ui/src/components/DashboardCard.vue
```diff
@@ -35,6 +33,7 @@ defineProps<{
   .tooltip {
     display: flex;
     justify-content: end;
+    cursor: default;
```
GoodVibezOnly:
```suggestion
    cursor: help;
```
Maybe?

Image

admin-console/ui/src/components/HealthIcon.vue
```diff
@@ -15,31 +15,33 @@ const presets: Record<HealthStatus, { severity: string; text: string; icon: stri
 }
 
 const { health } = defineProps<{ health: HealthStatus }>()
-const { icon, severity, text } = presets[health]
```
GoodVibezOnly: Nitpick: i prefer the destructured version but that is just personal preference

admin-console/ui/src/util/countries.ts
```diff
@@ -7,12 +7,15 @@ export function formatCountry(country: string) {
   try {
     return COUNTRY_NAMES.of(country)
   } catch {
+    if (country.toLowerCase() === 'aiida' || country === 'Unknown') {
```
GoodVibezOnly: Nitpick: maybe move to seperate variable `const aiidaCountries: ['aiida', 'Unknown']` and use like `aiidaCountries.includes(country.toLowerCase())`

admin-console/ui/src/util/countries.ts
```diff
@@ -21,5 +24,5 @@ export function countryFlag(countryCode: string) {
   }
   return [...countryCode]
     .map((char) => String.fromCodePoint(127397 + char.charCodeAt(0)))
-    .reduce((a, b) => `${a}${b}`)
+    .reduce((a, b) => `${a}${b}`, '')
```
GoodVibezOnly: is there ever the case we want this to return an array? 🤔 If no => just return single countryCode string value directly

admin-console/ui/src/views/DashboardView.vue
```diff
@@ -118,6 +121,9 @@ function getPermissionCountPerRegionConnector() {
       <header>
         <i class="pi pi-check-circle"></i>
         <h2>Permissions <span>Timeline</span></h2>
+        <RouterLink to="/permissions">
+          <i class="pi pi-chevron-right"></i>
```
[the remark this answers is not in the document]
GoodVibezOnly: [or in this case labelled-by :) ](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-labelledby)

admin-console/ui/src/components/HealthIcon.vue
re1: With this being a prop I think I would have to make this reactive with `computed` and `toRefs`, which feels like overkill.

admin-console/ui/src/components/SidebarNavigation.vue
```diff
@@ -45,6 +45,14 @@ import EddieIcon from '@/components/EddieIcon.vue'
 aside {
   padding: 1.75rem 1.25rem;
   background: var(--sidebar-background);
+
+  @media (width >= 80rem) {
+    position: sticky;
+    top: 0;
+    left: 0;
+    float: left;
+    height: 100dvh;
```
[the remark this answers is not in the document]
re1: I keep forgetting which viewport unit leads to overflows. I think both work _here_. I only want the sidebar to fill the whole height on desktop, while allowing the rest of the body to grow above viewport height. Anyway, this is only for desktop and the height shift is almost exclusively a mobile issue. Can change it to vh to avoid confusion.

admin-console/ui/src/views/DashboardView.vue
```diff
@@ -89,6 +89,9 @@ onMounted(async () => {
   permissions.value = await getPermissions()
   dataNeeds.value = await getDataNeeds()
   regionConnectors.value = await getRegionConnectors()
+})
+
+watchEffect(async () => {
```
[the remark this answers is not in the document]
re1: Good catch :+1: 

Should be fine here, since I only need regionConnectors.value to be tracked, not the regionConnectorHealth I am writing to.

admin-console/ui/src/views/DashboardView.vue
GoodVibezOnly: link for sim leads to a 404

Image
link should be changed to https://architecture.eddie.energy/framework/1-running/region-connectors/region-connector-simulation.html

not sure if this is just a local problem for me since my simulation region connector is called sim

GoodVibezOnly:
@re1 one personal gripe i have currently is that you are using `createMemoryHistory()` for vue router meaning i can not go back and forth between views via back button. I know that i can reach everything through the sidebar but im 100% sure people will still accidentally press the back button and then leave the admin console (which might lead to frustration).
I know this because that whats happens to me everyday when i use Loga 

This will not in anyway effect my approving this PR.

admin-console/ui/src/components/SidebarNavigation.vue
```diff
@@ -45,6 +45,14 @@ import EddieIcon from '@/components/EddieIcon.vue'
 aside {
   padding: 1.75rem 1.25rem;
   background: var(--sidebar-background);
+
+  @media (width >= 80rem) {
```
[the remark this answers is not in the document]
re1: Seems I was hasty with removing the wrapper. Brought it back.

fweingartshofer:
> Dead link check fails for a file I did not touch.

Seems like sonars netlify deployment broke.

re1:
> Let me check if it still works without issue in our Docker and Kubernetes setups.

Does work on my local machine. I hope Benji complains if something breaks (if anyone is using admin in K8s at all right now).

admin-console/ui/src/router/index.ts
```diff
@@ -1,11 +1,11 @@
...
+// SPDX-FileCopyrightText: 2024-2026 The EDDIE Developers <eddie.developers@fh-hagenberg.at>
 // SPDX-License-Identifier: Apache-2.0
 
-import { createMemoryHistory, createRouter } from 'vue-router'
+import { createRouter, createWebHistory } from 'vue-router'
 import DashboardView from '../views/DashboardView.vue'
 
 const router = createRouter({
-  history: createMemoryHistory(),
+  history: createWebHistory(),
```
GoodVibezOnly: <3

PR #2292, opened 2025-12-28
GH-2188: Add frontend component for DE-ETA region connector (ce.js)

bilal-sakhawat1:
This PR implements the frontend component required to integrate the DE-ETA region connector with the EDDIE button, following the Quickstart frontend guidelines. The custom element is exposed via the /ce.js endpoint and placed under the region connector’s public resources.

The component enables user interaction from the EDDIE UI and serves as the entry point for initiating permission requests. This completes the frontend–backend linkage required for region connector onboarding.

PR #2291, opened 2025-12-26
GH-2187: Initialize and configure DE-ETA region connector with permission request support

bilal-sakhawat1:
This PR implements the initial setup and core backend functionality for the region-connector-de-eta module in accordance with the Region Connector Quickstart documentation.

It includes the complete project initialization, required Spring configuration, and backend support for permission request creation, ensuring the region connector is buildable, discoverable, and operational within the EDDIE framework.

### Key Changes

**1. Initial Region Connector Setup**

- Created the region-connector-de-eta module following the prescribed project structure and conventions
- Added Gradle build configuration (build.gradle.kts), module metadata, and base package layout
- Ensured the connector can be built, started, and recognized by EDDIE as an independent and extensible region connector

**2. Spring Bean Configuration**

- Added Spring configuration to register all required runtime components, including:
- EventBus and Outbox infrastructure
- Permission request handlers
- Connection status and permission document providers
- JPA repositories and persistence configuration
- Integrated the connector with EDDIE’s event-driven architecture to support reliable event publishing and persistence

**3. Permission Request Creation API**

- Implemented backend support for creating permission requests as defined in the Quickstart guide
- Added a REST API endpoint and service layer to handle request validation and creation
- Published permission request events to the EventBus to initiate the standard EDDIE permission lifecycle

### Result

With this PR, the DE-ETA region connector is fully initialized, correctly wired into the EDDIE runtime, and capable of handling permission request creation, forming a solid foundation for frontend integration and further connector-specific functionality.

Issue #2187, opened 2025-11-18
Germany ETA+ Region Connector: Create a permission request at the permission administrators side


bilal-sakhawat1:
Set up the region-connector-de-eta module according to the Region Connector Quickstart documentation. The project structure, Gradle build configuration, and base metadata were created following the prescribed conventions. This establishes a runnable and discoverable region connector within the EDDIE framework and provides the foundation for further feature development.

Configured the required Spring beans for the DE-ETA region connector as defined in the Quickstart guide. This includes registration of the EventBus, Outbox, permission handlers, JPA repositories, and persistence configuration. The setup ensures proper event handling, persistence, and integration with EDDIE’s event-driven architecture.

Implemented backend support for creating permission requests in the DE-ETA region connector, following the Quickstart instructions. A REST API endpoint was added along with a service layer to validate and create permission requests and publish the corresponding domain events.

PR #2360, opened 2026-02-13
GH-2187 Remove non-existent eta npm dependencies from lock file

re1:
Introduced in #2291. Potentially breaks builds. Frontend will only be added in #2292.


GoodVibezOnly:
how did that even get in there? shouldnt running the build run pnpm install as well? :thinking:

Issue #2335, opened 2026-01-28
EDDIE Button: Subscribe to multiple permission requests


fweingartshofer:
## Short Description

Some region connectors support creating multiple permission requests with one creation request.
Allow the pop up to subscribe to the status updates and show them to the final customer.

## Acceptance Criteria

- [ ] Subscribe to multiple created permission requests
- [ ] Show status updates to the final customer

PR #2372, opened 2026-02-19
GH-2335 Fix permission status alerts for multiple data needs

re1:
Supporting multiple data needs in AIIDA since #2104 broke connection status messages for the AIIDA region connector, including previously working single data need scenarios. This PR does not consider other RCs and ignores #2356.

Commit 1 adds the bearer token for permission creation responses that include an array of `permissionIds` in addition to the existing logic for a single `permissionId`. The bearer is removed when setting the AIIDA code in the custom element.

Commit 2 adjusts the EDDIE Button to render separate status alerts per permission requests and to only change views if all permissions reached the same status.

:warning: Please note that I consider both changes to be quick fixes! Both our frontend and the permission request APIs require more work to achieve consistent user flows involving multiple data needs.

If the AIIDA UI tests are merged first, I will create a UI test handling multiple data needs. If not I will add one in a separate PR.

[Kooha-2026-02-19-17-19-49.webm](https://github.com/user-attachments/assets/b7985cbb-bf66-40b5-9577-110d832906eb)

The button only jumps if all permission requests have the same status to avoid information hiding.

Screenshot From 2026-02-19 17-21-53


tformatix (approved):
good job! :)

Issue #2346, opened 2026-02-03
AIIDA deployment fails on Windows


ewidl:
**Short description**

On Windows, the deployment of AIIDA via Docker Compose fails.
The reason is that the execution of the start-up scripts mounted into the containers for the `timescaledb` and `emqx` services fail.

**Steps to reproduce**

Deploy on Windows workstation:

``` bash
docker compose --profile dockerized-aiida up -d
```

**Logs**

Logs from container `aiida-timescaledb-1`:

``` bash
/usr/local/bin/docker-entrypoint.sh: running /docker-entrypoint-initdb.d/00-replace_password_with_env.sh
/usr/local/bin/docker-entrypoint.sh: line 185: /docker-entrypoint-initdb.d/00-replace_password_with_env.sh: cannot execute: required file not found
```

Logs from container `aiida-emqx-1`:

``` bash
/tmp/replace-password-with-env.sh: 2: 
/tmp/replace-password-with-env.sh: 5: 
/tmp/replace-password-with-env.sh: 6: set: Illegal option -
```

**Proposed solution**

On Windows, the shell scripts used for the `emqx` and `timescale` services are checked out with CRLF line endings.
There is already a `.gitattributes` file in the repo, I suggest adding the following line:
```
*.sh    text eol=lf
```

PR #2376, opened 2026-02-20
GH-2346 AIIDA deployment fails on Windows

Flo-Auzi:
- Change line ending for password replacement scripts in order for it to work in Docker Desktop for Windows


tformatix (approved):
looks valid :)

PR #2378, opened 2026-02-20
NOISSUE Make two tests more reliable

re1:
- Make `DataNeedRuleSet.dataNeedRules()` test match multiple data need rules instead of one
- Make `givenUnsupportedDataNeed_returnsDataNeedNotSupportedResult` not rely on the order of elements in a HashSet by ordering elements before printing them


fweingartshofer (approved):
The title of the PR is very general, maybe you can rephrase it to be a more specific

PR #2382, opened 2026-02-24
NOISSUE Remove AIIDA OAuth2 client superseded by resource server config

re1:
AIIDA previously relied on Spring being configured as an OAuth2 Client. This was because the Thymeleaf version of the AIIDA UI did not handle authentication itself. The Vue version however handles Keycloak authentication via a resource server, specifically using keycloak-js. With the complete removal of the old UI, there should no longer a need for Spring to act as a client server.


tformatix (approved):
🚀

Issue #2370, opened 2026-02-19
Publish CIM Maven package to Maven Central


tformatix:
## Short Description
Even if the repository is public, the Maven repository of GitHub still needs authentication (https://github.com/orgs/community/discussions/174091). @georghartner and I already tried that out - didn't work without providing a token.

@georghartner decided to provide the Maven package via Maven central (https://mvnrepository.com/repos/central). Keep in mind to update the CIM client library documentation.

## Acceptance Criteria
- [x] CIM Maven package is available in Maven central (wait until #2371 is merged)
- [x] Documentation updated -> Done in https://github.com/eddie-energy/eddie/pull/2387

PR #2384, opened 2026-02-25
GH-2370 publish cim maven package to maven central

fweingartshofer:
Still testing some things, but should be ready for review


tformatix (approved):
Very nice job! :)

PR #2386, opened 2026-02-25
GH-2144 Map additional fields to OBIS codes for FR data connector

Flo-Auzi:
- Map CCASN and CCAIN codes to respective OBIS codes
- Pass PRM instead of DataSourceId as mRID in CimStrategy v1_12
- Update StandardMode test for MicroTeleinfoV3Adapter


tformatix (commented):
looks very good :) two small tiny changes

PR #2390, opened 2026-02-26
NOISSUE core: unify connection status messages stream endpoint and provide helper to create the endpoint with permission ids

fweingartshofer:


region-connectors/shared/src/main/java/energy/eddie/regionconnector/shared/web/RestApiPaths.java
```diff
@@ -26,9 +30,22 @@ public final class RestApiPaths {
...
+    /**
+     * Utility method to create the url where connections status messages of specific permission requests can be consumed.
+     *
+     * @param permissionIds the ID of the permission requests in question.
+     * @return endpoint where connection status messages can be consumed.
+     */
+    public static URI connectionStatusMessagesStreamFor(String... permissionIds) {
+        return UriComponentsBuilder.fromUriString(CONNECTION_STATUS_STREAM_BASE)
+                                   // Enforce the varargs call
+                                   .queryParam("permission-id", (Object[]) permissionIds)
```
re1: Maybe Arrays.asList instead?

```suggestion
                                   .queryParam("permission-id", Arrays.asList(permissionIds))
```

Issue #2238, opened 2025-12-02
AIIDA: Topic for CIM Adapter in AIIDA UI leads to confusion


benjaminellmer:
## Short Description
Currently the UI shows the following info in the UI, for a CIM Adapter:

<img width="1292" height="1226" alt="Image" src="https://github.com/user-attachments/assets/cdc0c579-d46c-49f8-9d00-577933d9c56c" />

The problem here is, that AIIDA creates an ACL for the topic pattern: `aiida/0bba41b7-254b-4337-aa10-5b86bef640bb/+`, therefore the user is not authorized to exactly the topic `aiida/0bba41b7-254b-4337-aa10-5b86bef640bb` only for `/+`.
This problem could be solved by showing `aiida/0bba41b7-254b-4337-aa10-5b86bef640bb/+` or `aiida/0bba41b7-254b-4337-aa10-5b86bef640bb/data` in the UI.
However, I would prefer if we could rewrite the rule, so the user has access to the topic `aiida/0bba41b7-254b-4337-aa10-5b86bef640bb` as well (of course open for discussion).

## Acceptance Criteria
- [ ] Show a topic in the UI, to which the user is actually authorized to write

PR #2391, opened 2026-02-26
GH-2238 Topic for CIM Adapter in AIIDA UI leads to confusion

Flo-Auzi:
- Add /data suffix to topic name representation for UI

### DIFF
--- a/aiida/docs/1-running/data-sources/mqtt/cim/cim-data-source.md
+++ b/aiida/docs/1-running/data-sources/mqtt/cim/cim-data-source.md
@@ -17,20 +17,20 @@ Devices can send data to the MQTT broker of the AIIDA instance on the dedicated 
 ```json
 {
   "version": "1.0",
-  "registeredResourceMRID": {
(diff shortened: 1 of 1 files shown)


aiida/src/main/java/energy/eddie/aiida/models/datasource/mqtt/cim/CimDataSource.java
```diff
@@ -14,10 +14,17 @@
...
     protected CimDataSource() {}
 
     public CimDataSource(CimDataSourceDto dto, UUID userId) {
         super(dto, userId);
     }
+
+    @Override
+    protected String topicFormattedForUi() {
+        return super.topicFormattedForUi() + TOPIC_SUFFIX;
+    }
```
tformatix: Did you try it out? I think you need the same logic as for `OesterreichsEnergieDataSource`.

```suggestion
    @Override
    public void createAccessControlEntry() {
        var topic = TOPIC_PREFIX + id + TOPIC_SUFFIX;
        accessControlEntry = new MqttAccessControlEntry(id, topic);
    }
```

Issue #1713, opened 2025-04-02
Admin console: Allow sending retransmission requests to EDDIE


fweingartshofer:
## Short Description
Add the possibility the send retransmission requests to EDDIE.

## Acceptance Criteria
- [ ] It is possible to send retransmission requests to EDDIE

## Blockers
- [ ] _please list any blockers here that need to be finished so that this issue can be tackled_

PR #2410, opened 2026-03-10
GH-1713 Add retransmit button to admin console permissions view

re1:
Basic implementation for #1713. Lots of caveats because our admin console architecture aged poorly..

Added a new end date column to the status message table for passing the permission end date to the retransmission request.

While the issue does not specify a specific use-case, I want to do a second PR for making _from_ and _to_ dates of the retransmission request optional and configurable from the UI, easing flows like "send data up to now", "retransmit all data again", or "resend this time frame". Want to address lowish branch coverage / thin tests when the implementation is stable.

PR includes some related refactoring:
- Separate _permission start date_ and _status message creation date_ since the naming was already way too confusing.
- Reason is now passed to the frontend, which was already prepared for this in advance.

There are some other pain points I would like to note:
- Vue views could share fetched date → working on this separately.
- Permissions view does not show permission start and end yet → will do with other style changes since the number of columns starts to break usability.
- Errors resulting from the retransmission request are not sent to the frontend. Similar to termination, the request is delegated to the respective region connector and both results and errors are sent to a shared stream, not an immediate response. We would have to implement endpoints for the UI to subscribe to for that stream, which I consider out of scope for this PR → see #2412.
- I think that all the date columns, except the end date, _should_ be non-null and `TIMESTAMP WITH TIME ZONE` / `Instant`. However, I frequently find myself in situations where the date is null or empty, which I assume stems from specific RCs. As not to break things, they remain nullable and strings.

**Screencast**

Pointer is not visible, but the video shows (1) click on bulk retransmit with two selected and (2) click on first of selected.

https://github.com/user-attachments/assets/edb31dd8-12a4-48c9-9292-03d38df4e08d

### DIFF
--- a/admin-console/src/main/resources/db/migration/admin-console/V1_5__add_start_date_end_date_reason.sql.license
+++ b/admin-console/src/main/resources/db/migration/admin-console/V1_5__add_start_date_end_date_reason.sql.license
@@ -1,2 +1,0 @@
-SPDX-FileCopyrightText: 2026 The EDDIE Developers <eddie.developers@fh-hagenberg.at>
-SPDX-License-Identifier: Apache-2.0


re1:
> The retransmission request automatically requests everything for the whole timeframe of the permission, is that correct?

From start to yesterday or the permission end date if in the past. See PR description.

PR #2411, opened 2026-03-10
NOISSUE CIM: Shade Jackson dependencies to avoid Spring Boot 3.x conflicts with older Jackson 2 versions

fweingartshofer:
Work in progress, since there are more complications with shading jackson than I thought.


tformatix (commented):
as far as I understood, it looks good to me :)

tformatix (approved):

Issue #2409, opened 2026-03-10
CIM: Generate agnostic classes


tformatix:
## Short Description

`RawDataMessage`, `ConnectionStatusMessage`, and `OpaqueEnvelope` are currently implemented as manually created Java classes. Since they already exist in the schema repository https://github.com/eddie-energy/eddie/tree/main/cim/src/main/schemas/agnostic, they could also be provided through the CIM library.

Generating these agnostic classes as part of the CIM library would allow partners to use the generated classes directly, instead of maintaining their own Java implementations.

Therefore, the agnostic schemas should be included in the generation process so that the corresponding classes become part of the CIM library. The existing manually implemented Java classes can then be removed, as they would become redundant.

## Acceptance Criteria

- [x] `RawDataMessage`, `ConnectionStatusMessage`, and `OpaqueEnvelope` are generated and included in the CIM library 
- [x] The existing manually implemented Java classes are removed

PR #2416, opened 2026-03-11
GH-2409 Generate agnostic java classes in CIM lib

tformatix:


region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/permission/request/EdaDataSourceInformation.java
```diff
@@ -12,37 +12,21 @@
...
     private static final EdaRegionConnectorMetadata regionConnectorMetadata = EdaRegionConnectorMetadata.getInstance();
     @Column(length = 8)
     private final String dsoId;
 
     public EdaDataSourceInformation(String dsoId) {
         this.dsoId = dsoId;
+        this.countryCode = regionConnectorMetadata.countryCode();
+        this.regionConnectorId = regionConnectorMetadata.id();
+        this.meteredDataAdministratorId = dsoId;
+        this.permissionAdministratorId = dsoId;
```
fweingartshofer: The parent class should reduce complexity, by abstract away common fields, but cases like these, it creates more complexity, since instead of having one field, we get three fields with the same meaning in this scope.

fweingartshofer (commented):
I did only get through one tenth of the files, will try to continue reviewing later.
I was thinking about the purpose of the issue that you created, and I think the easier way that also preserves the API of the agnostic classes would be to move our java implementations to the cim package along with the schemas.

That way partners can use our java implementations or generate them on their own if they don't use java.
The only exception would be the `DataSourceInformation`, which would require an additional example implementation, but I think this issue can be solved easily.

PR #2438, opened 2026-04-07
NOISSUE Rename isAcknowledgementRequired and add AIIDA data need docs

tformatix:
Drop the is prefix from the acknowledgement flag across Java models, DB columns, JSON configs, Vue UI, JS components, locales, docs, and tests.


re1 (commented):
Changes seem fine, but what is the reason for removing the prefix? It is not stated in the PR description and no issue is linked.

re1 (approved):
Changes are not really related, but maybe wait for the E2E tests to finish. E2E runner fails to pick up tasks recently 🙄

fweingartshofer (approved):

PR #2445, opened 2026-04-16
fix(core): Disable raw data output by default but enable in dev

re1:
Using "fix" type to bump minor version despite being primarily docs since the new default technically produces a different behavior for the non-compose container.


fweingartshofer (approved):

re1:
Dead link is maintenance. Failing UI test is unrelated.

PR #2446, opened 2026-04-16
refactor: Replace old registrar pattern with a dynamic approach that wires providers and outbound connectors dynamically during start up

fweingartshofer:
Please enjoy this PR on a per commit basis.
I'm happy to discuss concerns and explain my approach directly if this PR seems to overwhelming or confusing.
I'm also open to other naming schemes.

Things to do if we are going forward with this approach:

- [x] Remove old registrars
- [ ] Add possibility to route messages from outbound connectors to region connectors -> Will not be done here since the PR is already very large
- [x] Document new approach in docs and in code comments

# Concerns

There might be some performance concerns since this approach uses reflection to scan spring beans and their methods to find the correct providers and consumers.
Keep in mind the code is only executed once after start up for each region connector right after all singletons have been initialized and when the context is refreshed.

# Comparison of Supported Features

I compared the output of the supported features endpoint before and after, and they are identical.


core/src/main/java/energy/eddie/core/services/MessageStreamHub.java
```diff
@@ -0,0 +1,55 @@
...
+    @SuppressWarnings("unchecked")
+    public synchronized <T> void registerProvider(Class<T> messageType, Supplier<Flux<T>> provider) {
+        var sink = (Sinks.Many<T>) sinksByType.computeIfAbsent(
+                messageType,
+                key -> Sinks.many().multicast().onBackpressureBuffer()
+        );
+
+        provider.get().subscribe(sink::tryEmitNext);
+
+        var receivers = receiversByType.get(messageType);
```
fweingartshofer, line 38: Check for double subscriptions.

core/src/main/java/energy/eddie/core/web/ConnectionStatusMessageController.java
```diff
@@ -3,39 +3,46 @@
...
+import reactor.core.publisher.Sinks;
 
 import java.util.List;
 
 import static energy.eddie.regionconnector.shared.web.RestApiPaths.CONNECTION_STATUS_STREAM_BASE;
 
 @RestController
 public class ConnectionStatusMessageController {
-    private final PermissionService permissionService;
+    private static final org.slf4j.Logger LOGGER = org.slf4j.LoggerFactory.getLogger(ConnectionStatusMessageController.class);
```
tformatix:
```suggestion
    private static final Logger LOGGER = LoggerFactory.getLogger(ConnectionStatusMessageController.class);
```

core/src/main/java/energy/eddie/core/web/ConnectionStatusMessageController.java
```diff
@@ -3,39 +3,46 @@
...
+                             .body(filtered);
+    }
+
+    @MessageStream(ConnectionStatusMessage.class)
+    void subscribe(Flux<ConnectionStatusMessage> flux) {
+        flux
+                .onErrorContinue((err, obj) -> LOGGER.warn(
+                        "Encountered error while processing connection status message",
+                        err))
+                .subscribe(messages::tryEmitNext);
```
tformatix, line 49: NITPICK: formatting is a little bit weird here

core/src/main/java/energy/eddie/core/message/streams/MessageStreamUtils.java
```diff
@@ -0,0 +1,33 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package energy.eddie.core.message.streams;
+
+import org.springframework.core.env.Environment;
+import reactor.core.publisher.Flux;
+
+import java.lang.reflect.Method;
+
+public class MessageStreamUtils {
```
tformatix, line 11: not a fan of util method/classes

tformatix (commented):
Very well done! Tested AIIDA and it works!

core/src/main/java/energy/eddie/core/message/streams/MessageStreamUtils.java
fweingartshofer: Same for me, but I didn't have a better solution that wouldn't be dogmatic.

re1: isReceiverMethod is only used once and areRawDataMessagesEnabled just wraps a property value. If you wanted, you could make them local to the StreamProviderAndSupplierRegistrar. In theory, RCs would only have providers anyway, but you could also duplicate isProviderMethod. Depends on if you favor the removal of the util class over some duplication. My rule of thumb is "fine up to 2 uses and up to 5 lines".

core/src/main/java/energy/eddie/spring/regionconnector/extensions/RegionConnectorSupportedFeatureExtension.java
```diff
@@ -116,4 +118,26 @@ private boolean hasBean(Class<?> clazz) {
         }
         return true;
     }
+
+    private boolean hasMessageStream(Class<?> clazz) {
+        if (context == null) {
```
re1: Can the application context actually be null? The constructor annotates the param with a supressed warning. To my knowledge, it can only be null with lazy init. Did not try it though. Same question for other null checks.

core/src/main/java/energy/eddie/core/message/streams/MessageStreamConfig.java
re1, line 1: We have our other smallish bean definitions collected in the CoreSpringConfig making this an outlier. Do not care too much.

re1 (commented):
Three smallish comments. Overall good work and a definite improvement. I confirmed that it runs with many RCs and OCs enabled and tested the data flow towards REST.

core/src/main/java/energy/eddie/core/message/streams/MessageStreamConfig.java
fweingartshofer: I wanted to keep it separate for the tests, since when executing spring tests it will initialize the bean contained in this config, which can lead to unexpected exceptions and behavior.
So I would rather keep it separate to not interfere with existing tests.

core/src/main/java/energy/eddie/core/message/streams/MessageStreamUtils.java
fweingartshofer: TL;DR: I want to keep the isXYZMethods in the same class for findability purposes and the areRawDataMessagesEnabled to have the property defined in one central place.

I want to have the `areRawDataMessagesEnabled` in one method to make possible refactoring changes to the environment value easier.
Furthermore, I would like to keep the isProviderMethod and isConsumerMethod in the same encasing unit to prevent confusion, since they are quiet similar and used for the same use case.

> fine up to 2 uses and up to 5 lines

I don't like those rules for code reviews, since they don't give reasons why it should be done this or that way.
If somebody wants to use them in their code I'm fine with it ( I use them too), but for reviews they don't make a lot of sense, since reviews should contain the why, and these rules never mention the why.

re1: You can choose your preference of course, I am just providing some inputs since you mentioned you are not a fan of the util class yourself. Do not care much personally, as I find it good enough and will likely never touch it :shrug:

> areRawDataMessagesEnabled to have the property defined in one central place

The property itself is already the "central place". Not sure about other's workflows but I always check (e.g. CTRL+SHIFT+F) on the property name first, and for typed config if I cannot find it.

> I would like to keep the isProviderMethod and isConsumerMethod in the same encasing unit

Totally understand, never argued against that.

> I don't like those rules for code reviews

It is _my_ rule of thumb. Just wanted to nudge your thinking in that direction assuming you already have an idea on the why. I am not writing an essay on AHA, DRY, Separation, Locality here.

core/src/main/java/energy/eddie/core/message/streams/MessageStreamConfig.java
re1: Fine with me, just assumed you would actually _want_ to load it for tests.

fweingartshofer: Not for all of them, because most tests are not whether the extension works, but on other things.
It does not make sense to me to force the tests to also **always** test this specific piece of code.

re1 (dismissed):

re1 (approved):

PR #2458, opened 2026-05-04
feat(sim): Produce raw data messages from simulation connector

re1:
Helps us explain raw data vs CIM data in the [EDDIE tutorial](https://github.com/eddie-energy/tutorial) and other documentation.

Thanks @fweingartshofer for the initial patch. I only adjusted the test to also cover the raw data messages.

### DIFF
--- a/region-connectors/region-connector-simulation/src/test/java/energy/eddie/regionconnector/simulation/engine/SimulationEngineTest.java
+++ b/region-connectors/region-connector-simulation/src/test/java/energy/eddie/regionconnector/simulation/engine/SimulationEngineTest.java
@@ -26,6 +26,7 @@ import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 import reactor.test.StepVerifier;
+import tools.jackson.databind.ObjectMapper;
 
 import java.time.Period;
 import java.time.ZoneOffset;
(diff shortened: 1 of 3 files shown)


region-connectors/region-connector-simulation/src/test/java/energy/eddie/regionconnector/simulation/engine/SimulationEngineTest.java
```diff
@@ -77,16 +79,18 @@ void testRun_withValidatedHistoricalData_emitsValidatedHistoricalDataAndPermissi
...
 
         // When
         var res = engine.run(scenario, metadata);
 
         // Then
         var ok = assertInstanceOf(SimulationStarted.class, res);
         ok.thread().join();
-        StepVerifier.create(streams.getValidatedHistoricalDataMarketDocumentsStream())
+        StepVerifier.create(Flux.merge(rawStream, validatedStream))
                     .expectNextCount(1)
```
fweingartshofer: Shouldn't this return 2 messages? One for the rawStream and one for the validatedStream?
But it seems like the tests are green, maybe I'm just confused.

re1: Created an extra test just for raw data to avoid this entirely

fweingartshofer (approved):

PR #2462, opened 2026-05-07
refactor(env): Merge sql and sh files for using env variable as db user password

re1:
Our env compose setups for EDDIE and AIIDA rely on replacing the password inside an SQL script with an env variable. I think this can be merged into one step where we just run the SQL from the shell script.

Wait for AIIDA UI test to confirm setup and communication still work.


re1:
E2E tests fail on Fluvius, not AIIDA.

tformatix (approved):
amazing - big improvement ^^

PR #2472, opened 2026-05-20
refactor(cim): Deserialize data source information to simple implementation by default

re1:
Not sure about which use-case we are currently tackling here. When mapping connection status messages using the client library they user would have to configure which implementation the data source information should map to → meh.

To my knowledge, we have no place where we benefit from not deserializing to `SimpleDataSourceInformation` by default.

Keeping this as a draft for now since I am not yet sure about the implications of this change.
Also, testing if something broke on the client library takes some time.


fweingartshofer (commented):
Is this still active? I'm happy with merging this PR.

fweingartshofer (approved):
Looks good, did you verify that the DataSourceInformation are correctly deserialized after saving it to the database, for example, via the REST outbound-connector?

re1:
> Looks good, did you verify that the DataSourceInformation are correctly deserialized after saving it to the database, for example, via the REST outbound-connector?

The EDDIE-side still works as expected. Tested connection status messages with the REST outbound connector. Reason this is a draft is that I did not bother to extensively test deserialization via the imported client library. To be honest, I will likely not put more brains into this until it becomes a problem again. Initially needed this for the tutorial. We _could_ merge this as a non-breaking change and hope that it helps in the long run :upside_down_face:

PR #2485, opened 2026-06-02
fix(aiida): Prevent custom select options from floating on scroll

re1:
## Bug

To reproduce
1. Resize window to require scroll on a page (e.g. > 2 data sources)
2. Scroll down a little
3. Click the select field
4. See screenshot

image

Tested in Firefox Developer Edition and Chromium.

## Fix

image

There is a separate issue for screen width < 694 where the `left: -1px` border fix hurts us. Not sure why it happens because there is no media query for this. Did not want to put too much effort into this, but maybe @GoodVibezOnly you have an idea?

image


fweingartshofer (approved):

GoodVibezOnly (approved):
tip top

PR #2486, opened 2026-06-02
feat(outbound-connectors): add PermissionCommand to Kafka and AMQP

### LINKED ISSUE #2405: AIIDA: Enable activation, deactivation, and schedule adjustment of AIIDA outbound data transmission by the eligible party
## User Story

---

## Context

Example:

```json
{
  "type": "outbound-aiida",
  "purpose": "purpose",
  "enabled": true,
  "duration": {
    "type": "relativeDuration",
    "start": "P0D",
    "end": "P3M"
  },
  "schemas": [
    "SMART-METER-P1-CIM"
  ],
}
```

---

## Goal

---

- activate transmission
- deactivate transmission

using control messages.

Example use cases:

---

Example:

---

Example:

If permission allows:

```
*/2 * * * * *
```

```
*/10 * * * * *
0 */1 * * * *
0 */5 * * * *
```

But not:

```
*/1 * * * * *
```

---

### Activate transmission

```json
{
  "action": "activate",
}
```

### Deactivate transmission

```json
{
  "action": "deactivate",
}
```

```json
{
  "action": "updateSchedule",
}
```

---

### AC1 — Activation control

### AC5 — State persistence

- active / inactive
- current effective schedule

---

## Technical Notes

```
data_need_runtime_state
```

Example fields:

```
data_need_id
active
effective_schedule
last_updated
```

---

## Benefits

- Tailored data exchange
- Lower infrastructure load

---

## Definition of Done

tformatix:


cim/src/main/java/energy/eddie/cim/agnostic/PermissionCommand.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Adjusts the transmission schedule for a permission.
+     *
+     * @param transmissionSchedule cron expression; effective schedule is capped at the data-need frequency
+     */
+    record UpdateSchedule(
+            String regionConnectorId,
+            UUID permissionId,
+            ZonedDateTime timestamp,
```
re1: What is the motivation behind the eligible party providing the timestamp in a command? Just traceability from their side?

core/src/main/java/energy/eddie/core/services/agnostic/PermissionCommandRouter.java
```diff
@@ -0,0 +1,77 @@
...
+
+    public void registerPermissionCommandConnector(PermissionCommandOutboundConnector permissionCommandOutboundConnector) {
+        LOGGER.info("Registering PermissionCommandOutboundConnector: {}",
+                    permissionCommandOutboundConnector.getClass().getName());
+        var subscription = permissionCommandOutboundConnector.getPermissionCommands()
+                                                             .subscribe(
+                                                                     this::route,
+                                                                     e -> LOGGER.error(
+                                                                             "Error in PermissionCommandRouter", e
+                                                                     ));
```
re1: Personal preference

```suggestion
        var subscription = permissionCommandOutboundConnector
                .getPermissionCommands()
                .subscribe(this::route, e -> LOGGER.error("Error in PermissionCommandRouter", e));
```

docs/1-running/outbound-connectors/outbound-connector-rest.md
```diff
@@ -38,6 +38,8 @@ The following table lists all the endpoints and its required scopes, if the sett
 |----------------------------------------------|---------------------------------------------|
 | `/agnostic/connection-status-messages`       | `agnostic-connection-status-messages:read`  |
 | `/agnostic/raw-data-messages`                | `agnostic-raw-data-messages:read`           |
+| `/agnostic/permission-command`               | `agnostic-permission-command:write`         |
+| `/agnostic/opaque-envelope`                  | `agnostic-opaque-envelope:write`            |
```
re1, line 42: Would you mind adding examples to `outbound-rest-requests.http` as well?

cim/src/main/java/energy/eddie/cim/agnostic/PermissionCommand.java
tformatix: You're right. I've been thinking about traceability, since these messages are retained in MQTT (and can therefore be retrieved at any time). But we could also use the MQTT timestamp.

That would anyways be complicated for the EP.

docs/1-running/outbound-connectors/outbound-connector-rest.md
tformatix: REST outbound connector are not yet enabled here - are still in TODO in #2405. Only added the Keycloak roles yet.

re1 (approved):
Good job!

PR #2495, opened 2026-06-03
feat(aiida): Add event duration and provide units in Watt for MinMax to OpenADR

re1:
Better aligns with its potential production use.

image


tformatix (approved):
great :)

fweingartshofer (approved):

PR #2497, opened 2026-06-03
feat(aiida): add allowed permission commands to data need

### LINKED ISSUE #2405: AIIDA: Enable activation, deactivation, and schedule adjustment of AIIDA outbound data transmission by the eligible party

tformatix:


re1 (commented):
This probably works against the entire PR, but we might want to (1) prepare for a scenario where additional commands are added, and (2) allow for more granular control of allowed commands. Instead of a global boolean flag, maybe we can pass a list of commands to allow. For example:

```json
"allowedControlCommands": [ "UPDATE_SCHEDULE", "SET_TRANSMISSION_ENABLED" ]
```

This approach means that when new commands are added, a new permission has to be created as well to support it. This can be seen as both a downside or a safety mechanism.

aiida/src/test/java/energy/eddie/aiida/streamers/mqtt/MqttStreamerTest.java
```diff
@@ -412,6 +414,61 @@ void verify_connectComplete_retransmitFailedMessages() throws MqttException {
...
+            streamer.messageArrived(EXPECTED_COMMAND_TOPIC, mockMessage);
+
+            // Then
+            assertThat(captor.getWarnLogs())
+                    .anyMatch(log -> log.contains("not in allowed permission commands"))
+                    .noneMatch(log -> log.contains("received SET_TRANSMISSION_ENABLED"));
+        }
+    }
+
+    private void setUpDataNeedWithPermissionCommands(Set<PermissionCommand.Action> allowedPermissionCommands) {
```
re1: Could this be combined with the helper below? Also, being in a test, the helper can just throw the original exception without the try-catch.

We could probably remove them entirely by just having a class field mock passed to the streamer creation in setup adding its mocking logic post-setup.

Tested this for the initial test before your changes:

Image

aiida/src/main/java/energy/eddie/aiida/streamers/mqtt/MqttStreamer.java
```diff
@@ -178,19 +178,35 @@ public void messageArrived(String topic, MqttMessage message) {
             return;
         }
 
+        var action = command.action();
+        if (action.requiresExplicitGrant() && !isPermissionCommandAllowed(action)) {
```
re1: Maybe merge the first check into the helper? My logic being "if the action does not require explicit grant, it is allowed".

```suggestion
        if (!isPermissionCommandAllowed(command.action())) {
```

re1 (dismissed):
Seems good now, just two minor suggestions.

aiida/src/main/java/energy/eddie/aiida/streamers/StreamerManager.java
```diff
@@ -82,56 +80,50 @@ public void createNewStreamer(Permission permission) throws IllegalArgumentExcep
...
+                .ifPresent(streamer -> streamer.setTransmissionEnabled(enabled));
+    }
+
+    /**
+     * Applies a new transmission schedule to the streamer of the passed permission by rebuilding its record flux with
+     * the new aggregation cadence.
+     */
+    public void updateSchedule(Permission permission) {
+        requireStreamer(permission.id())
+                .ifPresent(streamer -> buildFilteredFlux(permission).ifPresent(streamer::updateRecordFlux));
```
re1, line 116: Not sure how to tackle this and if we want to do this here. When we fail to update the schedule, the permission will have the new transmission schedule set but not send in that schedule. With the streamer still being active, there will be no indication to the user or EP that the updating failed. Closing the streamer feels even worse, as it will stop without indication. If the streamer for the permission ID does not exist (bad ID was sent) the failure will be silent.

re1 (dismissed):
Seems good. Some minor comment I am not sure of. Noticed that the PR does two migrations on the same system. We generally want to avoid this by grouping the migrations in one commit as not to pollute our code base with too many migrations.

aiida/src/main/java/energy/eddie/aiida/streamers/StreamerManager.java
tformatix: We will need to tackle those asynchronous issues in the future - will not be tackled in this PR.

re1 (approved):

PR #2498, opened 2026-06-03
feat(aiida): add transmission control logic to AIIDA

### LINKED ISSUE #2405: AIIDA: Enable activation, deactivation, and schedule adjustment of AIIDA outbound data transmission by the eligible party

tformatix:
image


aiida/src/main/java/energy/eddie/aiida/models/permission/Permission.java
```diff
@@ -116,6 +118,18 @@ public class Permission {
     @Nullable
     private InboundMessageFormat inboundMessageFormat;
 
+    @Column(name = "transmission_enabled", nullable = false)
+    @Schema(description = "Whether data transmission is currently enabled for this permission.")
+    @JsonProperty
+    private boolean transmissionEnabled = true;
```
tformatix, line 128: Add to UI

core/src/main/resources/application-dev.properties
re1, line 1: Seems accidental since unrelated

aiida/src/main/java/energy/eddie/aiida/streamers/StreamerManager.java
```diff
@@ -162,6 +152,29 @@ public void close() {
...
+        var userId = Objects.requireNonNull(permission.userId());
+        var dataSource = Objects.requireNonNull(permission.dataSource());
+
+        return aggregator.getFilteredFlux(allowedDataTags,
+                                          allowedAsset,
+                                          permissionExpirationTime,
+                                          transmissionSchedule,
+                                          userId,
+                                          dataSource.id());
+    }
```
re1: Not particularly pressing since it probably cannot happen in production, but all these null checks may throw uncatched NullPointerExceptions. We should find out what each missing value means and log them. The error handling around this seems a bit scattered overall.

Asset and data tags here are also not marked as nullable and do not really benefit from the check. Not sure Nullaway complains here.

aiida/src/main/java/energy/eddie/aiida/streamers/StreamerManager.java
```diff
@@ -82,56 +81,52 @@ public void createNewStreamer(Permission permission) throws IllegalArgumentExcep
...
+    public void setTransmissionEnabled(UUID permissionId, boolean enabled) {
+        requireStreamer(permissionId)
+                .ifPresent(streamer -> streamer.setTransmissionEnabled(enabled));
+    }
+
+    /**
+     * Applies a new transmission schedule to the streamer of the passed permission by rebuilding its record flux with
+     * the new aggregation cadence.
+     */
+    public void updateSchedule(Permission permission, CronExpression transmissionSchedule) {
```
re1: The transmission schedule is already available in the permission. Fine with passing it to be explicit about its use but not really necessary. Just want to make sure that it is intentional.

aiida/src/main/java/energy/eddie/aiida/models/permission/Permission.java
```diff
@@ -281,6 +308,7 @@ public void setMqttStreamingConfig(MqttStreamingConfig mqttStreamingConfig) {
     public void setDataNeed(AiidaLocalDataNeed dataNeed) {
         this.dataNeed = requireNonNull(dataNeed);
         this.serviceName = dataNeed.name();
+        this.transmissionSchedule = dataNeed.transmissionSchedule();
```
re1: We should stick to one approach for when the transmission schedule is null. Most of your code suggests that the transmission schedule remains null unless it is updated via the command. This makes it easy for us to check if the schedule was adapted, but needs us to handle null values on this field.  In that case, we should remove this line to keep it as null or keep its custom value.

The alternative would be for the schedule field to hold the value from the data need by default, which is implied on this line. If we want to take that approach, I suggest we add it for all existing permission on migration. I do however see little benefit aside from us being able to remove the `effectiveTransmissionSchedule` getter.

aiida/src/main/java/energy/eddie/aiida/models/permission/Permission.java
```diff
@@ -301,4 +329,22 @@ public void updateInboundMessageFormat(@NonNull InboundMessageFormat inboundMess
...
+    }
+
+    /**
+     * Returns the schedule that should be used to stream data for this permission: the permission's own
+     * transmission schedule if set, otherwise the data need's schedule as a fallback
+     */
+    public CronExpression effectiveTransmissionSchedule() {
+        return transmissionSchedule != null
+                ? transmissionSchedule
+                : requireNonNull(dataNeed).transmissionSchedule();
```
re1: This can throw a NullPointerException if no data need is set. Not something that would happen in production, but having this return null if neither data need nor transmission schedule is set is an easy hardening improvement.

aiida/ui/src/components/PermissionDetails.vue
```diff
@@ -30,6 +30,11 @@ const { permission, status } = defineProps<{
 }>()
 
 const { updatePermission } = usePermissionDialog()
+
+const effectiveTransmissionSchedule = computed(
+  () => permission.transmissionSchedule ?? permission.dataNeed.transmissionSchedule,
+)
```
re1: We could avoid this by passing `Permission#effectiveTransmissionSchedule` as a field when returning the permission instance as JSON.

aiida/src/main/java/energy/eddie/aiida/services/PermissionCommandService.java
```diff
@@ -0,0 +1,177 @@
...
+    private final PermissionCommandService self;
+
+    @Autowired
+    public PermissionCommandService(
+            PermissionRepository permissionRepository,
+            StreamerManager streamerManager,
+            Clock clock,
+            PermissionScheduler permissionScheduler,
+            AiidaEventPublisher aiidaEventPublisher,
+            @Lazy PermissionCommandService self
```
re1, line 55: Why is this reference needed? Might be worth a comment.

aiida/src/main/java/energy/eddie/aiida/streamers/StreamerManager.java
tformatix: You're right - I am using `effectiveTransmissionSchedule` now.

aiida/src/main/java/energy/eddie/aiida/models/permission/Permission.java
tformatix: Removed this. Kept the separate `transmissionSchedule` in the permission. This will help, if we want to track, if an schedule got changed at some time. Per default the schedule from the data need will be used.

aiida/src/main/java/energy/eddie/aiida/services/PermissionCommandService.java
tformatix: added a comment in the code

aiida/src/main/java/energy/eddie/aiida/streamers/StreamerManager.java
tformatix: Added error handling + logging

re1 (approved):
Seems fine 👍

PR #2502, opened 2026-06-08
fix(admin-console): Ensure Keycloak login directs relative to admin console module

re1:
Related to an issue where the Keycloak login redirect for the admin console does not work when behind a proxy. This does not add any configuration for ensuring the absolute path is correct. The URIs do not take absolute paths and I did not yet figure out a way to consistently use the public URL here.


tformatix (approved):
lgtm

fweingartshofer (approved):

PR #2504, opened 2026-06-09
feat(de-eta-region-connector): move credentials to separate table and remove on terminal states (GH-2198)

### LINKED ISSUE #2198: Germany ETA+ Region Connector: Remove credentials if possible

bilal-sakhawat1:
**GitHub Issue:** https://github.com/eddie-energy/eddie/issues/2198

---

## Description

OAuth credentials (`access_token`, `refresh_token`) were stored in `AcceptedEvent`, which
persists rows to the append-only `de_eta.permission_event` table. Because that table is
immutable, credentials could never be deleted — they remained in the database permanently
even after a permission was terminated, fulfilled, revoked, or became unfulfillable.

The EDDIE architecture requires credentials to be stored in a separate mutable table so
they can be removed when a permission's lifecycle ends. This PR introduces
`de_eta.permission_credentials` as that table, migrates existing data, wires all credential
reads/writes through a dedicated repository, and adds `PermissionRequestFinishedHandler`
to delete credentials on any terminal permission status.

---

### Behaviour

- Credentials are now stored exclusively in `de_eta.permission_credentials` and are never written to the event table going forward
- When a permission reaches any terminal status, its credentials are deleted immediately
- All three credential consumers (`AcceptedHandler`, `AccountingPointDataHandler`, `PollingService`) behave identically to before — only the credential source changes
- The one-shot refresh token exchange in `AccountingPointDataHandler` is intentionally not persisted (unchanged behaviour)


region-connectors/region-connector-de-eta/src/main/java/energy/eddie/regionconnector/de/eta/permission/handlers/AcceptedHandler.java
```diff
@@ -91,7 +96,18 @@ public void accept(AcceptedEvent event) {
...
 
-            fetchAndPublishData(pr, event.accessToken(), observation);
+            DePermissionCredentials creds = credentialsRepository.findByPermissionId(pr.permissionId())
+                    .orElse(null);
+            if (creds == null) {
+                LOGGER.atWarn()
+                      .addArgument(pr::permissionId)
+                      .log("No credentials found for permission {}, marking UNABLE_TO_SEND");
+                observation.stop();
+                commitSafely(event.permissionId(), PermissionProcessStatus.UNABLE_TO_SEND);
```
fweingartshofer: The status should be unfulfillable, since you cannot fulfill the permission request without the credentials.
`UNABLE_TO_SEND` should only be used when initially sending the permission request to the final customer and the permission administrator is not available.
For example, when trying to retrieve the access token from the OAuth Server using the code in the authorization code grant.

region-connectors/region-connector-de-eta/src/main/java/energy/eddie/regionconnector/de/eta/permission/handlers/AccountingPointDataHandler.java
```diff
@@ -87,10 +93,16 @@ public void accept(AcceptedEvent event) {
             return;
         }
 
-        String refreshToken = event.refreshToken().orElse(null);
+        DePermissionCredentials creds = credentialsRepository.findByPermissionId(pr.permissionId())
+                .orElse(null);
+        if (creds == null) {
+            LOGGER.warn("No credentials found for permission {}", pr.permissionId());
+            commitSafely(pr.permissionId(), PermissionProcessStatus.UNABLE_TO_SEND);
```
fweingartshofer: Same here.

region-connectors/region-connector-de-eta/src/main/java/energy/eddie/regionconnector/de/eta/permission/request/events/AcceptedEvent.java
fweingartshofer, line 1: Use the `SimpleEvent` instead, since they are almost the same.
Don't forget to update the database migration, since the dtype `DeAcceptedEvent` won't be available anymore.

fweingartshofer (approved):

PR #2512, opened 2026-06-12
fix(de-eta-region-connector): drop and recreate permission request view in V1_4 migration (GH-2511)

### LINKED ISSUE #2511: Fix V1_4 DE-ETA migration: use DROP VIEW + CREATE VIEW instead of CREATE OR REPLACE VIEW
**Describe the bug**
The DE-ETA region connector migration `V1_4__add_permission_credentials_table.sql` fails to apply on PostgreSQL, which aborts application startup. The migration uses `CREATE OR REPLACE VIEW` to redefine `de_eta.eta_permission_request` while removing the `access_token` and `refresh_token` columns. PostgreSQL only allows `CREATE OR REPLACE VIEW` to add new trailing columns, so it cannot drop or reorder existing ones, and the statement is rejected with SQL state 42P16.

After `V1_3`, the view ends with the columns `..., cause, access_token, refresh_token, latest_meter_reading`. `V1_4` redefines it as `..., cause, latest_meter_reading`, which both drops two columns and reorders the tail. That is why PostgreSQL reports "cannot drop columns from view".

This affects every environment that uses PostgreSQL (local Docker, CI, and deployments). It is not Windows specific.

Affected file: `region-connectors/region-connector-de-eta/src/main/resources/db/migration/de-eta/V1_4__add_permission_credentials_table.sql` (line 26).

**To reproduce**
Describe the steps to reproduce the behavior:
1. Configure the DE-ETA connector against a fresh PostgreSQL database.
2. Build the application and image, then run `docker compose up -d` from the `env/` directory.
3. Watch the `eddie` container logs during startup (`docker compose logs eddie`).
4. See the Flyway error and the container exit with code 1.

**Expected behavior**
The `V1_4` migration applies cleanly, the `de_eta.eta_permission_request` view is recreated without the credential columns, and the application starts normally.

**Logs**

```
Failed to execute script V1_4__add_permission_credentials_table.sql
SQL State  : 42P16
Message    : ERROR: cannot drop columns from view
Location   : db/migration/de-eta/V1_4__add_permission_credentials_table.sql (line 26)
Caused by: org.postgresql.util.PSQLException: ERROR: cannot drop columns from view
```

**Screenshots**
Not applicable.

**Data Needs Configuration**
Not relevant for this bug.

**Additional context**
Proposed fix: replace the `CREATE OR REPLACE VIEW` with an explicit `DROP VIEW` followed by `CREATE VIEW`, so the column set is allowed to change. The preceding `INSERT ... FROM de_eta.eta_permission_request` still reads the credential columns and runs before the drop, so credential migration is preserved.

```sql
-- Postgres cannot drop or reorder existing columns via CREATE OR REPLACE VIEW,
-- so the view must be dropped and recreated to remove the credential columns.
DROP VIEW de_eta.eta_permission_request;
CREATE VIEW de_eta.eta_permission_request AS
SELECT DISTINCT ON (permission_id) permission_id,
    ... -- unchanged SELECT body
```

Since this migration has never successfully applied on PostgreSQL, editing it in place does not create a Flyway checksum conflict.

bilal-sakhawat1:
GH-2511 Fix V1_4 permission request view migration
---
## Description

The `V1_4` migration redefined the `de_eta.eta_permission_request` view to remove the `access_token` and `refresh_token` columns, but it did so with `CREATE OR REPLACE VIEW`. PostgreSQL only allows that statement to append new trailing columns, so it cannot drop or reorder existing ones. As a result the migration failed with SQL state 42P16 ("cannot drop columns from view") and the application could not start against PostgreSQL.

This change drops the view and recreates it, which is the only way to change its column set in PostgreSQL. The insert that copies existing credentials into `de_eta.permission_credentials` still reads from the view and runs before the drop, so no data is lost during the migration.

## Behaviour

- The `V1_4` migration now applies successfully on PostgreSQL.
- The recreated `de_eta.eta_permission_request` view no longer exposes `access_token` or `refresh_token`.
- Existing credentials for accepted permissions are copied into `de_eta.permission_credentials` before the view is dropped, so the migration preserves them.


fweingartshofer (dismissed):
You cannot change existing migrations

re1 (dismissed):
As stated in the issue description, the migration was never valid and therefore has to be changed post-merge. Please take this as a reminder to always test your application with Flyway enabled before committing a migration. Due to us being short-handed, we do not test run all migrations ourselves on review. We would usually catch such issue through our E2E tests, but we currently do not have ETA enabled for those.

bilal-sakhawat1:
> You cannot change existing migrations

@re1 
I added a new migration file, now what should I do? keep the old one or use this new migration file?

re1:
> > You cannot change existing migrations
> 
> I added a new migration file, now what should I do? keep the old one or use this new migration file?

There was a misunderstanding with Florian's comment. For this scenario, you indeed have to modify the existing migration file, since Flyway would not be able to migrate with a broken migration file in-between migrations. Therefore, you should revert to your previous version where you edited the existing migration.

Not "critical", since the ETA region connector is not used by any partner (to my knowledge), but better to not have a broken migration in the first place.

re1 (approved):
Confirmed myself locally that the migration now works, thanks for the quick fix!

bilal-sakhawat1:
> Confirmed myself locally that the migration now works, thanks for the quick fix!

@re1 Thanks alot

Issue #2394, opened 2026-03-04
AIIDA: FCA 1.1 - Forward inbound messages via AIIDA outbound data need


tformatix:
**Priority:** P2

**As a** flexibility service provider or energy community operator
**I want** to create an AIIDA outbound Data Need that forwards all incoming messages from a topic
**So that** messages received via AIIDA inbound permissions can automatically be redistributed to other systems.

## Description
The system should allow configuration of an aiida-outbound Data Need that subscribes to all inbound messages from a specified topic and republishes them to a defined outbound destination.

This enables topic-level forwarding / relay functionality within AIIDA.

## Acceptance Criteria
- [ ] An AIIDA outbound Data Need can be configured to subscribe to an inbound topic.
- [ ] All messages received via the inbound topic are forwarded to the outbound destination.
- [ ] Forwarded messages remain unchanged (transparent relay).
- [ ] Forwarding must support:
  - [ ] Kafka
  - [ ] REST endpoint
  - [ ] other AIIDA connectors (if configured)
  - [ ] metadata
- [ ] Errors during forwarding must be logged.
- [ ] Message throughput must not block inbound processing.

re1:
>  Forwarded messages remain unchanged (transparent relay)

As mentioned in the assumptions of the PR for this, the metadata on the messages is adapted to use the outbound permission ID and data need ID to trace the result. The inbound permission ID and data need ID are omitted. This is because we assume the forwarding outbound data need to be from a separate eligible party than the one that sent the messages in the first place. That eligible party would not have any permission or data need of the given IDs.

The envisioned data exchange looks like this.

```mermaid
sequenceDiagram
    box EP1
        participant EP1 as Eligible Party 1
        participant REST1 as REST Outbound
        participant RC1 as AIIDA RC
    end
 
    participant AIIDA as AIIDA
     
    box EP2
        participant RC2 as AIIDA RC
        participant KAFKA2 as Kafka Outbound
        participant EP2 as Eligible Party 2
    end
 
    EP1->>REST1: POST opaque/min-max
    REST1->>RC1: POST opaque/min-max
    RC1->>AIIDA: inbound message
    Note over AIIDA: replace headers
    AIIDA->>RC2: forwarded message
    RC2->>KAFKA2: forwarded message
    EP2->>KAFKA2: opaque/min-max
```

Issue #2476, opened 2026-05-27
AIIDA: InboundDataSource Refactoring


Flo-Auzi:
## Short Description
The InboundDataSources and their corresponding elements such as the aggregator and record data models are not yet implemented so they can be easily adapted for future features and use cases. This should be changed here in a first step by adding a separate InboundAggregator.

## Acceptance Criteria
- [ ] Current logic is changed such that InboundRecords are stored inside a separate Sink that is exposed via a separate Aggregator

PR #2513, opened 2026-06-15
GH-2394 Allow selection of AIIDA inbound permissions as data sources

re1:
This PR introduces the following changes towards implementing the forwarding of inbound messages in AIIDA for #2394.
1. Eligible parties may request inbound schemas in outbound data needs.
2. Customers may select inbound data sources (presented as permissions) for outbound permissions.
3. Inbound records are relayed to the MQTT topic of the outbound permission and schema.

The changes only implement the AIIDA side: requesting, selection, and forwarding.
I want to split the EDDIE side into its own PR to keep the total number of changes manageable.

A couple assumptions were made when working on these features, and some questions arose that I would like to list here.

## Assumptions

 - Presence of an inbound schema means a forwarding data need requiring the selection of an inbound data source
 - Allow only the selection of data sources able to produce one of the requested schemas
 - Only one data source can be selected per permission
 - EP receiving the forwarded messages does not need the original metadata
	 - EP would not request forwarded messages if they were sending the messages themselves. Therefore, metadata can be replaced when forwarded to match the forwarding permission.

## Questions

- [x] Allow adding data needs with both inbound and outbound schemas or fail creation?
- [x] Should the data source be able to forward one requested schema or all?
	- Outbound schemas → select outbound data source
	- Inbound schemas → select inbound data source that can forward it
	- Mix of inbound and outbound → select either inbound or outbound
	- Both inbound and outbound → select one of each
- [x] Throw when trying to set a data source with the wrong asset type, do we care?
- [x] Allow the user to name their permissions? Data need name as default.
	- Most relevant for inbound, but might also be useful for outbound permissions.
- [x] How to handle revoke inbound permission?
	- Revoke outbound permissions using the inbound data source as well, or
	- only allow revoke if no outbound permissions use its data source (current implementation)?

## Screencasts

Selecting an inbound data source for an outbound permission

https://github.com/user-attachments/assets/a48e2670-f2b5-4a24-a704-7f38ba0e17a5

Handling revocation

https://github.com/user-attachments/assets/5bceea29-da9c-48f9-a797-589680d5a775

Downloading the latest forwarded record

https://github.com/user-attachments/assets/5dff87c2-e384-4b68-8d54-84e53af844bd

### DIFF
--- a/aiida/src/main/java/energy/eddie/aiida/repositories/PermissionRepository.java
+++ b/aiida/src/main/java/energy/eddie/aiida/repositories/PermissionRepository.java
@@ -7,35 +7,24 @@ import energy.eddie.aiida.models.permission.Permission;
 import energy.eddie.aiida.models.permission.PermissionStatus;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.data.jpa.repository.Query;
-import org.springframework.data.repository.query.Param;
 
 import java.util.List;
+import java.util.Set;
 import java.util.UUID;
 
 public interface PermissionRepository extends JpaRepository<Permission, UUID> {
-    /**
-     * Returns all permission objects for the given userId, sorted by their grantTime descending.
-     *
-     * @return A list of permissions, sorted by grantTime descending, i.e. the permission with the newest grantTime is the first item.
-     */
     List<Permission> findByUserIdOrderByGrantTimeDesc(UUID userId);
 
-    /**
-     * Returns a list of all active permission, i.e. all permissions, whose status is either {@link PermissionStatus#ACCEPTED} or {@link PermissionStatus#WAITING_FOR_START} or {@link PermissionStatus#STREAMING_DATA}.
-     *
-     * @return A list of permissions that have the status {@link PermissionStatus#ACCEPTED} or {@link PermissionStatus#WAITING_FOR_START} or {@link PermissionStatus#STREAMING_DATA}.
-     */
-    @Query("SELECT p FROM Permission p WHERE p.status IN (energy.eddie.aiida.models.permission.PermissionStatus.WAITING_FOR_START, energy.eddie.aiida.models.permission.PermissionStatus.STREAMING_DATA)")
-    List<Permission> findAllActivePermissions();
+    List<Permission> findByStatusIn(Set<PermissionStatus> statuses);
 
     @Query("""
             SELECT p.permissionId
             FROM Permission p
             WHERE p.dataSource.id = :dataSourceId
               AND p.dataNeed.type = energy.eddie.dataneeds.needs.aiida.OutboundAiidaDataNeed.DISCRIMINATOR_VALUE
-              AND p.status IN (energy.eddie.aiida.models.permission.PermissionStatus.WAITING_FOR_START, energy.eddie.aiida.models.permission.PermissionStatus.STREAMING_DATA)
+              AND p.status IN (:statuses)
             """)
-    List<UUID> findActiveOutboundPermissionIdsByDataSourceId(UUID dataSourceId);
+    List<UUID> findOutboundByDataSourceIdAndStatus(UUID dataSourceId, Set<PermissionStatus> statuses);
 
     @Query("""
             SELECT p
@@ -43,12 +32,8 @@ public interface PermissionRepository extends JpaRepository<Permission, UUID> {
             WHERE p.userId = :userId
               AND p.dataNeed.type = energy.eddie.dataneeds.needs.aiida.InboundAiidaDataNeed.DISCRIMINATOR_VALUE
               AND p.dataSource IS NOT NULL
-              AND p.status IN (
-                    energy.eddie.aiida.models.permission.PermissionStatus.FETCHED_MQTT_CREDENTIALS,
-                    energy.eddie.aiida.models.permission.PermissionStatus.WAITING_FOR_START,
-                    energy.eddie.aiida.models.permission.PermissionStatus.STREAMING_DATA
-              )
+              AND p.status IN (:statuses)
             ORDER BY p.grantTime DESC
             """)
-    List<Permission> findActiveInboundPermissionsByUserId(@Param("userId") UUID userId);
+    List<Permission> findInboundByUserIdAndStatus(UUID userId, Set<PermissionStatus> statuses);
 }
--- a/aiida/src/main/java/energy/eddie/aiida/repositories/PermissionRepository.java
+++ b/aiida/src/main/java/energy/eddie/aiida/repositories/PermissionRepository.java
@@ -26,4 +26,13 @@ public interface PermissionRepository extends JpaRepository<Permission, UUID> {
      */
     @Query("SELECT p FROM Permission p WHERE p.status IN (energy.eddie.aiida.models.permission.PermissionStatus.WAITING_FOR_START, energy.eddie.aiida.models.permission.PermissionStatus.STREAMING_DATA)")
     List<Permission> findAllActivePermissions();
+
+    @Query("""
+            SELECT p.permissionId
+            FROM Permission p
+            WHERE p.dataSource.id = :dataSourceId
+              AND p.dataNeed.type = energy.eddie.dataneeds.needs.aiida.OutboundAiidaDataNeed.DISCRIMINATOR_VALUE
+              AND p.status IN (energy.eddie.aiida.models.permission.PermissionStatus.WAITING_FOR_START, energy.eddie.aiida.models.permission.PermissionStatus.STREAMING_DATA)
+            """)
+    List<UUID> findActiveOutboundPermissionIdsByDataSourceId(UUID dataSourceId);
 }
--- a/aiida/src/main/java/energy/eddie/aiida/repositories/PermissionRepository.java
+++ b/aiida/src/main/java/energy/eddie/aiida/repositories/PermissionRepository.java
@@ -18,13 +18,13 @@ public interface PermissionRepository extends JpaRepository<Permission, UUID> {
     List<Permission> findByStatusIn(Set<PermissionStatus> statuses);
 
     @Query("""
-            SELECT p.permissionId
+            SELECT p
             FROM Permission p
             WHERE p.dataSource.id = :dataSourceId
               AND p.dataNeed.type = energy.eddie.dataneeds.needs.aiida.OutboundAiidaDataNeed.DISCRIMINATOR_VALUE
               AND p.status IN (:statuses)
             """)
-    List<UUID> findOutboundByDataSourceIdAndStatus(UUID dataSourceId, Set<PermissionStatus> statuses);
+    List<Permission> findOutboundByDataSourceIdAndStatus(UUID dataSourceId, Set<PermissionStatus> statuses);
 
     @Query("""
             SELECT p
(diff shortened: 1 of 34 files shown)


aiida/src/main/resources/db/aiida/migration/V9_1__link_inbound_data_source_to_owner_permission.sql
```diff
@@ -0,0 +1,22 @@
+--  SPDX-FileCopyrightText: 2026 The EDDIE Developers <eddie.developers@fh-hagenberg.at>
+--  SPDX-License-Identifier: Apache-2.0
+
+ALTER TABLE data_source_mqtt_inbound
+    ADD COLUMN permission_id uuid;
```
re1, line 5: We _should_ set this to NOT NULL as well once updated, but I frequently ended up in scenarios where with inbound data sources without a permission. I assume this happens when a permission is revoked, but the data source deletion fails. I fear that we will not be able to make a migration where all inbound data sources get an ID without deleting all existing data sources that currently do not have one. Therefore, I decided against the NOT NULL constraint as the least invasive migration.

aiida/src/main/java/energy/eddie/aiida/repositories/PermissionRepository.java
```diff
@@ -26,4 +27,28 @@ public interface PermissionRepository extends JpaRepository<Permission, UUID> {
...
+            SELECT p
+            FROM Permission p
+            WHERE p.userId = :userId
+              AND p.dataNeed.type = energy.eddie.dataneeds.needs.aiida.InboundAiidaDataNeed.DISCRIMINATOR_VALUE
+              AND p.dataSource IS NOT NULL
+              AND p.status IN (
+                    energy.eddie.aiida.models.permission.PermissionStatus.FETCHED_MQTT_CREDENTIALS,
+                    energy.eddie.aiida.models.permission.PermissionStatus.WAITING_FOR_START,
+                    energy.eddie.aiida.models.permission.PermissionStatus.STREAMING_DATA
+              )
```
tformatix: I am feeling a little bit uncomfortable on using status checks here - but it is okay for me.

aiida/src/main/java/energy/eddie/aiida/services/PermissionService.java
```diff
@@ -95,7 +99,8 @@ public PermissionService(
     @Transactional
     public Permission revokePermission(
             UUID permissionId
-    ) throws PermissionNotFoundException, PermissionStateTransitionException, UnauthorizedException, InvalidUserException {
+    ) throws PermissionNotFoundException, PermissionStateTransitionException, UnauthorizedException,
+             InvalidUserException, InboundDataSourceInUseException {
```
tformatix: Formatting on throwing exception in this class is not super-pretty.

aiida/src/main/java/energy/eddie/aiida/services/PermissionService.java
```diff
@@ -295,6 +303,11 @@ public List<Permission> getAllPermissionsSortedByGrantTime() throws InvalidUserE
         return permissionRepository.findByUserIdOrderByGrantTimeDesc(currentUserId);
     }
 
+    public List<Permission> getSelectableInboundPermissions() throws InvalidUserException {
```
tformatix: Is the purpose clear with that method name? Selectable for what?

aiida/src/main/java/energy/eddie/aiida/services/PermissionService.java
```diff
@@ -473,12 +486,47 @@ private void markPermissionAsUnfulfillable(Permission permission) throws Permiss
         throw new PermissionUnfulfillableException(permission.id());
     }
 
+    private void validateDataSourceSelection(Permission permission, UUID dataSourceId)
+            throws DataSourceNotFoundException, IncompatibleDataSourceException {
+        var dataSource = dataSourceService.dataSourceByIdOrThrow(dataSourceId);
```
tformatix, line 473: Not directly related to your PR but we're exposing data sources to other users with that call. We should authorize this service method.

aiida/src/main/java/energy/eddie/aiida/services/PermissionService.java
```diff
@@ -473,12 +486,47 @@ private void markPermissionAsUnfulfillable(Permission permission) throws Permiss
...
+            Objects.equals(inboundDataSource.permission(), permission)) {
             aiidaEventPublisher.publishEvent(new InboundPermissionRevokeEvent(dataSource.id()));
             permission.setDataSource(null);
         }
     }
+
+    private void validateInboundDataSourceNotInUse(Permission permission) throws InboundDataSourceInUseException {
+        var dataSource = permission.dataSource();
+        if (dataSource instanceof InboundDataSource inboundDataSource &&
+            Objects.equals(inboundDataSource.permission(), permission)) {
```
tformatix: this condition is checked twice in this class &rarr; extract to method

tformatix (commented):
Skimmed through already - looks very good. 

Didn't try it out yet and will continue with a more detailed review after the questions are resolved in the refinement.

re1:
The questions have been addressed with Georg in today's refinement:

> Allow adding data needs with both inbound and outbound schemas or fail creation?

Allow. While it does not make much sense semantically, there is no harm in allowing it.

> Should the data source be able to forward one requested schema or all?

The data source selected for a permission should support at least one requested schema.

> Throw when trying to set a data source with the wrong asset type, do we care?

Yes, the asset type of the selected data source should match the requested asset type. #2520 was defined to implement this for both inbound and outbound permissions.

> EP can currently only specify a single asset type -> refactor to list?

Yes, instead of a single asset type, it should be possible to define a list of options. When given an empty list, any asset type should be allowed. #2519 has been defined to implement this for EDDIE and AIIDA.

> Allow the user to name their permissions? Data need name as default?

Yes. A custom name can be saved in AIIDA to display to the customer. #2521 was defined for this purpose.

aiida/src/main/java/energy/eddie/aiida/services/PermissionService.java
re1: Fun-fact, IntelliJ will reformat them to wrap long lines when using a refactoring action, but put them on a single line when removing custom line breaks on format. Will fix it here but not make an effort to fix it for existing code.

aiida/src/main/java/energy/eddie/aiida/repositories/PermissionRepository.java
re1: Unfortunately you cannot pass in a static collection here if not as a parameter of the repository method. Having the status defined in this way helps remove the need for passing the parameters.

Passing them as parameters moves the filtering semantics into the service layer:

```java
public static final Set<PermissionStatus> ACTIVE = Set.of(WAITING_FOR_START, STREAMING_DATA);

public List<Permission> getActiveInboundPermissions() throws InvalidUserException {
    var currentUserId = authService.getCurrentUserId();
    return permissionRepository.findInboundPermissionsByUserIdAndStatus(currentUserId, ACTIVE);
}
```

Also fine with this version, but I would then apply this approach for the whole PermissionRepository.

aiida/src/main/resources/db/aiida/migration/V9_1__link_inbound_data_source_to_owner_permission.sql
re1, line 1: Merging this after #2476, so the naming will be correct.

aiida/src/main/java/energy/eddie/aiida/repositories/PermissionRepository.java
re1: Added a commit with the parameter approach and relocating the "type of status" semantics to the enum so you can check it out.

aiida/src/main/java/energy/eddie/aiida/utils/PermissionExpiredRunnable.java
```diff
@@ -53,9 +53,7 @@ public void run() {
               .log("Will expire permission permission {}, running {} ms too late compared to target time (negative number would be too early)");
 
         // safeguard if e.g. a revocation operation could not properly cancel this runnable before it runs
-        if (!(permission.status() == PermissionStatus.ACCEPTED ||
-              permission.status() == PermissionStatus.WAITING_FOR_START ||
-              permission.status() == PermissionStatus.STREAMING_DATA)) {
+        if (!permission.status().isActive()) {
```
tformatix, line 56: only an idea: instead of negating we could `isInactive`

aiida/src/main/java/energy/eddie/aiida/services/PermissionService.java
```diff
@@ -316,7 +324,7 @@ public Permission findById(UUID permissionId) throws PermissionNotFoundException
     @Transactional
     public void onApplicationEvent(@NonNull ContextRefreshedEvent event) {
         // fields of permissions are loaded eagerly to avoid n+1 select problem in loop
-        List<Permission> allActivePermissions = permissionRepository.findAllActivePermissions();
+        var allActivePermissions = permissionRepository.findByStatusIn(ACTIVE);
```
tformatix: maybe a little bit misleading - here active is defined differently as the list in `isActive`. 🤔

aiida/src/test/java/energy/eddie/aiida/services/InboundPermissionRelayServiceTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+        relayService.subscribeToInboundRecords();
+        var inboundPayload = """
+                {
+                  "regionConnectorId":"aiida",
+                  "permissionId":"inbound-permission",
+                  "connectionId":"inbound-connection",
+                  "dataNeedId":"inbound-data-need",
+                  "messageId":"inbound-message",
+                  "timestamp":"2026-06-19T12:00:00Z",
```
tformatix, line 85: there is no `timestamp` and `messageId` in the opaque envelope

tformatix (dismissed):
nice job :)

aiida/src/test/java/energy/eddie/aiida/services/InboundPermissionRelayServiceTest.java
re1: There is not? :thinking:

image

aiida/src/main/java/energy/eddie/aiida/utils/PermissionExpiredRunnable.java
re1: Was playing around with that idea and IntelliJ complained as well. However, I found it more confusing the other way around since `isActive` makes it very clear that it has the same values as the `ACTIVE` set exposed from that same enum. Do not really care much though

aiida/src/test/java/energy/eddie/aiida/services/InboundPermissionRelayServiceTest.java
tformatix: true haha - thought we deleted that

aiida/src/main/java/energy/eddie/aiida/services/PermissionService.java
re1: I assumed from the naming and comment of the previous `findAllActivePermissions` and the code flow that the `ACCEPTED` state being missing from the query was accidental, or its meaning got lost. Read into the PR adding it a little and could not find any comment or other explanation on why this change was made. With `ACCEPTED` having been removed with changes to the streamer, I assume that this is/was for the case that no data source was set for an accepted permission. Just to be super safe, added a separate collection of states with waiting and streaming.

Good case for a "lying comment" as we recently talked about in the clean code session :sweat_smile:

tformatix (approved):

PR #2518, opened 2026-06-16
feat(simulation-region-connector): Extend scenarios with an interaction block for termination requests

### LINKED ISSUE #2517: Simulation Region Connector: Add Termination Request Interaction
## Short Description
When executing a test scenario it should be possible to await a termination request and only move on with the scenario once the request was received.

## Acceptance Criteria
- Test scenarios are extended with an interaction step that awaits termination requests

fweingartshofer:
The simulation-region-connector now supports scenarios with a termination interaction step that allows scenarios to await termination requests before moving on to the next steps

### DIFF
--- a/region-connectors/region-connector-simulation/src/main/java/energy/eddie/regionconnector/simulation/engine/constraints/TerminationFollowedByConstraint.java
+++ b/region-connectors/region-connector-simulation/src/main/java/energy/eddie/regionconnector/simulation/engine/constraints/TerminationFollowedByConstraint.java
@@ -1,0 +1,28 @@
+// SPDX-FileCopyrightText: 2026 The EDDIE Developers <eddie.developers@fh-hagenberg.at>
+// SPDX-License-Identifier: Apache-2.0
+
+package energy.eddie.regionconnector.simulation.engine.constraints;
+
+import energy.eddie.cim.agnostic.PermissionProcessStatus;
+import energy.eddie.regionconnector.simulation.engine.constraints.results.ConstraintOk;
+import energy.eddie.regionconnector.simulation.engine.constraints.results.ConstraintResult;
+import energy.eddie.regionconnector.simulation.engine.constraints.results.ConstraintViolation;
+import energy.eddie.regionconnector.simulation.engine.steps.Model;
+import energy.eddie.regionconnector.simulation.engine.steps.StatusChangeStep;
+import energy.eddie.regionconnector.simulation.engine.steps.TerminationInteractionStep;
+
+public class TerminationFollowedByConstraint implements StructuralConstraint {
+    @Override
+    public ConstraintResult violatesConstraint(Model current, Model next) {
+        if (!(current instanceof TerminationInteractionStep) || next == null) {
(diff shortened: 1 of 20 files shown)


region-connectors/region-connector-simulation/src/main/java/energy/eddie/regionconnector/simulation/engine/constraints/TerminationFollowedByConstraint.java
re1, line 1: Generally, the permission could remain in the TERMINATED state, can it not? I assume from this that you want to specifically want the simulation connector to use the REQUIRES_EXTERNAL_TERMINATION branch? If so, why not change it automatically to that state as you do in `WaitForTerminationStep#onTermination`?

Image

fweingartshofer: No,, in the first if statement you can see that it is ok if the termination is followed by nothing: `next == null`

re1: Stupid of me. I was focusing too much on the naming and did not follow the code flow correctly.

re1 (dismissed):

region-connectors/region-connector-simulation/src/main/java/energy/eddie/regionconnector/simulation/engine/constraints/TerminationFollowedByConstraint.java
fweingartshofer: Haha, no worries, I also did a double take.

region-connectors/region-connector-simulation/src/main/java/energy/eddie/regionconnector/simulation/engine/steps/runtime/WaitForTerminationStep.java
```diff
@@ -21,23 +20,19 @@ public WaitForTerminationStep(Duration waitFor) {
...
+            var res = ctx.documentStreams()
+                         .getTerminationStream()
+                         .any(permissionId -> permissionId.equals(ctx.permissionId()))
+                         .timeout(waitFor)
+                         .block();
+            if (Boolean.TRUE.equals(res)) {
+                return List.of(new StatusEmissionStep(PermissionProcessStatus.TERMINATED));
+            }
+        } catch (RuntimeException ignored) {
+            // No Op
```
re1, line 34: Just unsure about the comment here. The No Op here is actually us not reacting to the specific exceptions, but deferring throwing an exception to the next statement. Might work out here, since both the result being "false" and the `TimeoutException` from `timeout` can be interpreted as "never received a termination message". A `RuntimeException` from `block` is probably not realistic without having introduced a bug.

Maybe its just me re-reading this multiple times to confirm your intention with ignoring the exception, but a small actual comment stating why we ignore the exception might help.

re1 (approved):

PR #2523, opened 2026-06-18
fix(core): Fix bug where management APIs could be accessed from the public port when context-path is given

fweingartshofer:
Fix a bug where if the context path is set via `server.servlet.context-path`, the `InternalEndpointsFilter` would not match any management paths and would make all of them available via the public port instead of the management port. The reason for this is that the `InternalEndpointsFilter` does not take the context path into account when matching paths.


core/src/main/java/energy/eddie/core/ManagementApiConfig.java
```diff
@@ -57,7 +59,7 @@ public ManagementApiConfig(
 
     @Bean
     public ServletWebServerFactory servletContainer() {
-        Connector connector = new Connector(TomcatServletWebServerFactory.DEFAULT_PROTOCOL);
+        Connector connector = new Connector(TomcatWebServerFactory.DEFAULT_PROTOCOL);
```
fweingartshofer, line 62: Unrelated, but fixes a sonar warning.

benjaminellmer (approved):

PR #2524, opened 2026-06-19
fix(aiida): Sonar issues

Flo-Auzi:
- Resolve small sonar issues arisen from inbound data source refactoring

### DIFF
--- a/aiida/src/test/java/energy/eddie/aiida/aggregator/OutboundAggregatorTest.java
+++ b/aiida/src/test/java/energy/eddie/aiida/aggregator/OutboundAggregatorTest.java
@@ -15,7 +15,6 @@ import energy.eddie.api.agnostic.aiida.ObisCode;
 import nl.altindag.log.LogCaptor;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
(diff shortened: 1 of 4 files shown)


aiida/src/test/java/energy/eddie/aiida/aggregator/OutboundAggregatorTest.java
```diff
@@ -74,9 +74,10 @@ class OutboundAggregatorTest {
     @BeforeEach
     void setUp() {
         StepVerifier.setDefaultTimeout(Duration.ofSeconds(2));
+        now = Instant.parse("2026-06-19T12:00:00Z");
```
tformatix: maybe not call it now then (as it is not now anymore)

tformatix (commented):
tiny issue - otherwise it looks good to me :)

aiida/src/test/java/energy/eddie/aiida/aggregator/OutboundAggregatorTest.java
Flo-Auzi: Renamed to "start" as the other timestamps reference this as the valid starting time.

fweingartshofer (commented):
I think the OutboundAggregatorTest broke.

tformatix (dismissed):

Flo-Auzi:
> I think the OutboundAggregatorTest broke.

Yes, it got inconsistent because we now created two AiidaRecords with the exact same timestamp instead of the previous sequential Instant.now(). -> fixed

tformatix (approved):

PR #2534, opened 2026-06-26
GH-2533 rest outbound connector allow sse with xml

### LINKED ISSUE #2533: REST Outbound Connector: Allow SSE with XML
## Short Description

The REST Outbound Connector offers a number of SSE endpoints, but all of them only support JSON, which is problematic since the CIM is an XML first schema in EDDIE.
This should be changed so XML can also be consumed from an SSE endpoint

## Acceptance Criteria
- [x] SSE endpoints support xml

fweingartshofer:


outbound-connectors/outbound-rest/src/main/java/energy/eddie/outbound/rest/web/cim/v1_04/CimSwagger.java
```diff
@@ -135,6 +135,73 @@ public interface CimSwagger {
...
+                                                  <ns5:reason.code>999</ns5:reason.code>
+                                                </ns5:TimeSeries>
+                                              </ns5:MarketDocument>
+                                            </ns5:VHD_Envelope>
+                                            """
+                            )
+                    )
+            )
+    )
+    ResponseEntity<Flux<String>> validatedHistoricalDataMdSSEXML();
```
re1, line 152: Example does not work in Swagger UI:
http://localhost:9090/outbound-connectors/rest/swagger-ui/index.html#/CIM%20v1.04%20Documents/GET%20validated%20historical%20data%20market%20documents%20v1.04

Image

outbound-connectors/outbound-rest/src/main/java/energy/eddie/outbound/rest/web/EventStream.java
re1, line 1: Had a couple of pain points related to this class.

Since you are already on the refactoring, we can make this a helper for both XML and JSON SSE endpoints that can hold the buffering and content-type semantics.

I personally find the name too vague, and it reads more like a struct. Went with `EventStreamHelper` in my patch.

I do not see a point in making the `X-Accel-Buffering` header a constant. The `SSE` class holding the constant also seems pointless to me. It was long overdue to extract the header adding logic from the individual controllers, so with a central point it is not even a question. Even here the benefit is just lookup on (re)moving at some point, and one could just CTRL+F for the header name.

Since I wanted to test my refactoring I already have a patch ready. Use what you want.
[move_sse_response_building_to_helper.patch](https://github.com/user-attachments/files/29387281/move_sse_response_building_to_helper.patch)

I consider this out of scope, but in my opinion the controller methods should not have to care about the format of the payload or conversions and instead just specify which content types it can produce. Did not have the patience yet to implement a converter and also will not ask this from you. Spring might have the tooling needed to achieve this easily, but I did not find them as no one seems to use SSE with XML payloads.

outbound-connectors/outbound-rest/src/main/java/energy/eddie/outbound/rest/web/EventStream.java
```diff
@@ -0,0 +1,57 @@
...
+                             .body(flux.mapNotNull(this::toXml));
+    }
+
+    @Nullable
+    private String toXml(Object o) {
+        try {
+            var bytes = this.serde.serialize(o);
+            return new String(bytes, StandardCharsets.UTF_8);
+        } catch (SerializationException e) {
+            LOGGER.warn("Error converting to XML: {}", o, e);
```
re1, line 63: Not sure if logging and returning null is the right call here. The endpoint could return an error payload instead. Spring probably does this for errors on JSON conversion, and maybe it even has SSE-specific utility or this.

re1 (changes requested):

outbound-connectors/outbound-rest/src/main/java/energy/eddie/outbound/rest/web/EventStream.java
fweingartshofer: The point is to have the same header that is littered all over the module in one constant.
Not going to use the patch, since I really dislike helper classes, as they promote procedural code.

outbound-connectors/outbound-rest/src/main/java/energy/eddie/outbound/rest/web/cim/v1_04/CimSwagger.java
fweingartshofer: Seems to be an issue with the swagger library not with the code.
Probably because it does not know the media type.

outbound-connectors/outbound-rest/src/main/java/energy/eddie/outbound/rest/web/EventStream.java
re1: My suggestions and patch conflict with neither of your points. I would have already called your initual SSE and XmlSSE classes "helpers" and only applied that interpretation to the naming. If you do not want to call it Helper, call it Mapper? Changing its name does not change the behavior.

outbound-connectors/outbound-rest/src/main/java/energy/eddie/outbound/rest/web/EventStream.java
fweingartshofer: This cannot be done, since this is a server sent events endpoint, and the error might occur later on, after the initial response with the http code was already sent.

I don't know about a better option here.
This is definitely an internal server error, but apart from logging it I don't know a better solution.

re1: Did not mean a Java Exception with a Spring ExceptionHandler or HTTP error. Server-sent events themselves can hold information if the current message is an error. 

For normal messages the SSE response would be

```
event: message
data: ... XML or JSON payload
```

The event type could also be an error that the client can listen to

```
event: error
data: Error converting the message to XML
```

Not sure about the exact field names, but something like that.

I would assume that Spring has some mechanism for this for JSON where it auto-converts. To customize the behaviour you could return a ServerSentEvent object instead of String directly. See https://www.baeldung.com/spring-server-sent-events#2-making-use-of-the-serversentevent-element.

However, I am not sure if this level of detail/effort actually helps us. Would leave this up to you since you are closer to its actual use.

outbound-connectors/outbound-rest/src/main/java/energy/eddie/outbound/rest/web/agnostic/AgnosticController.java
```diff
@@ -114,7 +122,14 @@ public ResponseEntity<List<ConnectionStatusMessage>> connectionStatusMessagesJso
     public ResponseEntity<Flux<RawDataMessage>> rawDataMessagesSSE() {
         return ResponseEntity.ok()
                              // Tell reverse proxies like Nginx not to buffer the response
-                             .header("X-Accel-Buffering", "no").body(agnosticConnector.getRawDataMessageStream());
+                             .header(SSEEndpoints.X_ACCEL_BUFFERING, "no")
+                             .body(agnosticConnector.getRawDataMessageStream());
```
re1: Not using your helper here?

outbound-connectors/outbound-rest/src/main/java/energy/eddie/outbound/rest/web/agnostic/AgnosticSwagger.java
```diff
@@ -66,6 +66,21 @@ public interface AgnosticSwagger {
...
 
+
+    @Operation(
+            operationId = "GET ConnectionStatusMessage stream",
+            summary = "Get ConnectionStatusMessage stream",
+            description = "Get all new ConnectionStatusMessages as Server Sent Events",
+            responses = @ApiResponse(
+                    responseCode = "200",
+                    content = @Content(
+                            mediaType = "application/vnd.eddie.energy.sse+xml",
```
re1: Why not use the constant we defined for the media type?

outbound-connectors/outbound-rest/src/main/java/energy/eddie/outbound/rest/web/EventStream.java
```diff
@@ -0,0 +1,67 @@
...
+import reactor.core.publisher.Flux;
+
+import java.nio.charset.StandardCharsets;
+
+import static org.springframework.http.MediaType.TEXT_EVENT_STREAM_VALUE;
+
+@Component
+public class SSEEndpoints {
+    public static final String EVENT_STREAM_XML_VALUE = "application/vnd.eddie.energy.sse+xml";
+    public static final MediaType EVENT_STREAM_XML = MediaType.parseMediaType(EVENT_STREAM_XML_VALUE);
```
re1, line 24: Could move this to the test, personal preference.

outbound-connectors/outbound-rest/src/main/java/energy/eddie/outbound/rest/web/EventStream.java
```diff
@@ -0,0 +1,67 @@
...
+import org.springframework.stereotype.Component;
+import reactor.core.publisher.Flux;
+
+import java.nio.charset.StandardCharsets;
+
+import static org.springframework.http.MediaType.TEXT_EVENT_STREAM_VALUE;
+
+@Component
+public class SSEEndpoints {
+    public static final String EVENT_STREAM_XML_VALUE = "application/vnd.eddie.energy.sse+xml";
```
re1, line 23: Why the change from `text/event-stream+xml`? Cannot really wrap my head around why `application/vnd.eddie.energy.sse+xml` would work better. Neither are "recognized" media types, but I would argue that the previous version was closer to what kind of data we send. With just the `+xml` specifying that we do some conversion. `text/event-stream` itself does not make any assumption about the format of the payload. I guess that Spring defaults to JSON because it is the most common format. To my knowledge, applications typically use different endpoints or query params to indicate the format of the payload of the event stream instead of the `Accept` header. Us using it is more of a shortcut.

outbound-connectors/outbound-rest/src/main/java/energy/eddie/outbound/rest/web/SSEEndpoints.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.nio.charset.StandardCharsets;
+
+import static org.springframework.http.MediaType.TEXT_EVENT_STREAM_VALUE;
+
+@Component
+public class SSEEndpoints {
+    public static final String EVENT_STREAM_XML_VALUE = "application/vnd.eddie.energy.sse+xml";
+    public static final MediaType EVENT_STREAM_XML = MediaType.parseMediaType(EVENT_STREAM_XML_VALUE);
+    public static final String X_ACCEL_BUFFERING = "X-Accel-Buffering";
```
re1: There should be no reason for this to be public if we already extract its use into this class.

outbound-connectors/outbound-rest/src/main/java/energy/eddie/outbound/rest/web/SSEEndpoints.java
```diff
@@ -0,0 +1,67 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(SSEEndpoints.class);
+    private final XmlMessageSerde serde;
+
+    public SSEEndpoints() throws SerdeInitializationException {
```
re1: Not a fan of the name. Saying that the class manages any endpoints or endpoint information is a long-shot and I feel it does not indicate what it actually does. If you want to avoid `Util` or `Helper` you could make it a `Mapper` since it "maps" payloads from an object to a string or a `ServerSentEventBuilder` conveying its building the response entity for a given stream.

outbound-connectors/outbound-rest/src/main/java/energy/eddie/outbound/rest/web/SSEEndpoints.java
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * Converts a flux of any type to a stream of XML documents, suitable for SSE endpoints.
+     * It changes the <code>Content-Type</code>-Header to <code>text/event-stream</code>, since otherwise clients are not able to receive any events.
+     * Drops objects that cannot be serialized to an XML document.
+     *
+     * @param flux the source for the events
+     * @param <T>  the type of the source of the events
+     * @return a stream of XML documents
+     */
+    public <T> ResponseEntity<Flux<String>> xml(Flux<T> flux) {
```
re1: Type information should not be needed?

```suggestion
    public ResponseEntity<Flux<String>> xml(Flux<?> flux) {
```

outbound-connectors/outbound-rest/src/main/java/energy/eddie/outbound/rest/web/SSEEndpoints.java
```diff
@@ -0,0 +1,67 @@
...
+        return ResponseEntity.ok()
+                             // Tell reverse proxies like Nginx not to buffer the response
+                             .header(X_ACCEL_BUFFERING, "no")
+                             .header("Content-Type", TEXT_EVENT_STREAM_VALUE)
+                             .body(flux.mapNotNull(this::toXml));
+    }
+
+    public <T> ResponseEntity<Flux<T>> json(Flux<T> flux) {
```
re1: Why does XML deserve a Javadoc comment but not JSON :(

outbound-connectors/outbound-rest/src/test/java/energy/eddie/outbound/rest/web/AgnosticControllerTest.java
```diff
@@ -75,6 +76,29 @@ void connectionStatusMessageSSE_returnsMessages() {
                     .verifyComplete();
     }
 
+    @Test
+    @DirtiesContext
+    void connectionStatusMessageSSEXML_returnsMessages() {
```
re1, line 81: The test might not be that helpful, as it only checks if the endpoint exists, returns values, and that it streams. It does lack the rather important aspect of checking if the payload is in the right format, which is why we have these endpoints in the first place. Not that I find any of these tests particularly relevant.

re1 (commented):
Made some small comments from stuff that I previously had in my patch that I did not seem to have communicated clearly.

re1 (commented):
Only mentioned this briefly in a previous review comment, but we should still consider if a custom content type is even the right implementation here. Generally, the `text/event-stream` media type does not indicate any payload format. It could be JSON as in our case, but also an arbitrary string, base64 encoded, XML, etc. 

Maybe the "better" implementation would be to add a "format" parameter that can be one of JSON or XML, defaulting the parameter to JSON so we keep our existing API. This would remove the need for a custom media type and the need to manually update the response content type that currently cannot match the request Accept header.

The reason I give this PR so much attention is that any message endpoint we add to the REST outbound connector will have to use this pattern, meaning it will be used a lot in the code and endpoint definitions. It might also be hard to change our approach once some partner adopted it.

outbound-connectors/outbound-rest/src/main/java/energy/eddie/outbound/rest/web/EventStream.java
re1: Read up on the RFC you mentioned. The suffix is indeed only allowed if the media type including that suffix is specifically recognized, rendering my suggestion of keeping it invalid.

https://datatracker.ietf.org/doc/html/rfc6838#section-4.2.8

outbound-connectors/outbound-rest/src/main/java/energy/eddie/outbound/rest/web/SSEEndpoints.java
fweingartshofer: I try to avoid most -er names, as they have been long regarded as bad practice, unless they are required by some convention: 

- https://steve-yegge.blogspot.com/2006/03/execution-in-kingdom-of-nouns.html
- https://www.yegor256.com/2015/03/09/objects-end-with-er.html
- https://arxiv.org/html/2403.17430v1

fweingartshofer:
> Only mentioned this briefly in a previous review comment, but we should still consider if a custom content type is even the right implementation here. Generally, the `text/event-stream` media type does not indicate any payload format. It could be JSON as in our case, but also an arbitrary string, base64 encoded, XML, etc.

The `ACCEPTS` and `Content-Type` header are exactly for those situations.
I know that other projects may choose to send other formats, but we chose those formats for those content-types, so I see no reason to not do it that way.

outbound-connectors/outbound-rest/src/test/java/energy/eddie/outbound/rest/web/WebTestConfig.java
```diff
@@ -14,4 +15,9 @@ public class WebTestConfig {
     public RequestedContentTypeResolver requestedContentTypeResolver() {
         return new RequestedContentTypeResolverBuilder().build();
     }
+
+    @Bean
+    public EventStreamResource sseEndpoints() throws SerdeInitializationException {
```
re1: Remnant from rename. Not sure why the testing works, since there would be no bean mapping for the injected dependency. Was the bean used before?

```suggestion
    public EventStreamResource eventStreamResource() throws SerdeInitializationException {
```

re1 (commented):
Might not be in scope, but I realized the agnostic endpoints each state that they only produce JSON, which is not the case anymore.

Image

outbound-connectors/outbound-rest/src/main/java/energy/eddie/outbound/rest/web/EventStream.java
re1: Clarified in person that this is (1) a rare case we do not need to address and (2) the client this is used in and those we expect might have trouble processing error payloads while not benefitting from them.

outbound-connectors/outbound-rest/src/main/java/energy/eddie/outbound/rest/web/SSEEndpoints.java
re1: Clarified outside GitHub to avoid ping-pong comments. Naming was adjusted and stated as intentional.

outbound-connectors/outbound-rest/src/main/java/energy/eddie/outbound/rest/web/SSEEndpoints.java
re1: Wildcard type was kept due to concerns of it being picked up by Sonar as a code smell. Was unable to reproduce in IDE or CLI, but could be due to my caching.

outbound-connectors/outbound-rest/src/main/java/energy/eddie/outbound/rest/web/SSEEndpoints.java
re1: Adding the buffering header alone was not deemed relevant enough to add a comment. Changing the content type header and dropping payloads that failed transforming was.

outbound-connectors/outbound-rest/src/test/java/energy/eddie/outbound/rest/web/AgnosticControllerTest.java
re1: Checking payload transformation was not deemed relevant enough to extend this or another test.

re1 (approved):
Fine with me now. Quality Gate does not reflect actual testing and duplication levels. "Missing" coverage is on recurring endpoint definitions where formats are tested as one to avoid testing Spring magic. "Duplication" is us intentionally separating CIM version endpoints.

PR #2535, opened 2026-06-30
chore: Add @re1 as AIIDA code owner

re1:
Suggested by @tformatix as I am now responsible for implementing flexible connection agreements into AIIDA and will probably take on more AIIDA tasks in the future.

Placed first alphabetically, not in priority.


tformatix (approved):
🫡

PR #2541, opened 2026-07-02
GH-2395 Pass FCA context and meter ID to AIIDA permission requests

### LINKED ISSUE #2395: AIIDA: FCA 2.1 - Tag FCA inbound/outbound permissions
**Author**: @georghartner 

**Priority:** P3

**As a** system administrator
**I want** to mark exactly ONE or none inbound permission and ONE or none outbound permission as "FCA"; more - per AIIDA account - should not be possible. 
**So that** AIIDA can operate in a dedicated FCA context.

## Description
AIIDA must support a special FCA context flag to identify permissions that relate to Flexible Connection Agreements.

## Acceptance Criteria
- [ ] Permissions (or DataNeeds) support a boolean flag `for_fca`.
- [ ] The system allows:
  - [ ] maximum one inbound permission with `for_fca=true`
  - [ ] maximum one outbound permission with` for_fca=true`
- [ ] Attempting to create a second FCA permission results in an error.
- [ ] GUI indicates FCA permissions clearly.
- [ ] API validation enforces uniqueness.

re1:
Resolves #2395 by implementing the following functionality:
1. Add contexts field to AIIDA data needs to inform about the FCA context
2. Pass a meter ID set on the EDDIE button in the permission request for verification
3. Show an FCA label to highlight FCA permissions
4. Fail adding a second FCA inbound or outbound permission
5. Add an E2E test confirming the FCA context semantics

See the comments in #2395 for questions and refinements of the FCA context semantics.

The PR also includes some testability improvements and an error propagation fix.

The highlighting of FCA permissions was moved to #2546 as it needs further clarification and possibly more refactoring.

Two **implementation questions**:
- Should the meter ID be tracked in the permission events? It is not needed to function so it would only be there for tracing.
- Make specific exceptions following an unfulfillable state extend `UnfulfillableException` to compact throws list and exception handler registration in controller?

The `contexts` field is only added to AIIDA data needs with an `AiidaContext` enum type while keeping in mind that we might want to generalize it for non-AIIDA data needs later (CESU).

As stated in recent refinements we assume that the physical meter ID can be used to map connection agreement point. The eligible party will usually know the meter ID of the customer and can pass it in the permission request. The meter ID can be used for permission without FCA context as well. For example to limit data source selection for outbound permissions.

I added two separate FCA data needs as not to conflict with existing non-FCA development and test cases.

The provided FCA context is visible in the form of an "FCA" label on the permission and the optional presence of a "meter ID".

image

## Screencast

Meter ID

https://github.com/user-attachments/assets/48ddc2e7-105e-4cb3-b26c-5cd3859939ac

Duplicate meter ID

https://github.com/user-attachments/assets/1174ce60-d586-40ec-a00d-2863cbdbc3c9


admin-console/ui/src/components/DataNeedForm.vue
```diff
@@ -256,7 +262,7 @@ async function submitForm() {
       <template v-if="form.type === 'inbound-aiida' || form.type === 'outbound-aiida'">
         <div class="field">
           <label for="asset">Asset</label>
-          <Select id="asset" v-model="form.asset" :options="ASSETS" required />
+          <Select id="asset" v-model="form.asset" :options="Array.from(ASSETS)" required />
```
GoodVibezOnly, line 265: why array from? 🤔

GoodVibezOnly (commented):
Frontend looks good 👍

admin-console/ui/src/components/DataNeedForm.vue
re1: TypeScript complains because I appended the "as const" to reuse the constant as a type and now it needs some extra nudging to use it as an array :/

aiida/src/main/java/energy/eddie/aiida/repositories/PermissionRepository.java
```diff
@@ -36,4 +37,21 @@ AND p.status IN (:statuses)
...
+            SELECT CASE WHEN COUNT(p) > 0 THEN true ELSE false END
+            FROM Permission p
+            JOIN p.dataNeed dn
+            JOIN dn.contexts c
+            WHERE p.meterId = :meterId
+              AND p.status IN (:statuses)
+              AND dn.type = :dataNeedType
+              AND c = :context
+            """)
+    boolean existsByDataNeedTypeAndContextAndMeterIdAndStatusIn(
```
re1: We might want to limit this check to the AIIDA user. If not limited, a malicious actor could find out if a meter ID is connected to an FCA. Not sure if this is particularly useful and outweighs the benefit of our ability to check if the meter ID is already used in the AIIDA instance. If limited, any user could create an FCA for the meter ID.

aiida/src/main/resources/db/aiida/migration/V9_3__add_meter_id_to_permission.sql
re1, line 1: Kept the migrations separate intentionally. They have very different semantics that can be clearly separated per commit which might allow for easier revert/undo/refactor later with the only downside being the additional migration file. Can merge if found necessary.

aiida/src/test/java/energy/eddie/aiida/errors/GlobalExceptionHandlerTest.java
```diff
@@ -158,26 +161,6 @@ void givenInvalidInboundPermissionException_returnsBadRequest() {
         assertEquals(message, responseBody.get(ERRORS_PROPERTY_NAME).getFirst().message());
     }
 
-    @Test
-    void givenPermissionUnfulfillableException_returnsBadRequest() {
```
re1, line 162: Not sure if we need to test every single exception handler assignment. I consider this config level and Spring-magic. Can add the tests for the new exceptions if found relevant.

aiida/src/main/java/energy/eddie/aiida/repositories/PermissionRepository.java
re1: After some thought I adjusted the method to also check for the user, just to be safe.

aiida/src/main/java/energy/eddie/aiida/services/PermissionService.java
```diff
@@ -430,42 +430,67 @@ private Permission updatePermissionWithDetails(
...
-        var dataNeed = permission.dataNeed();
-        return dataNeed != null && isValidDataNeedType(dataNeed.type());
+    private void validatePermissionStartInTheFuture(Permission permission) throws PermissionStartInThePastException {
+        var start = requireNonNull(permission.startTime()).atZone(AIIDA_ZONE_ID);
+        var today = LocalDate.now(AIIDA_ZONE_ID).atStartOfDay(AIIDA_ZONE_ID);
+        if (start.isBefore(today)) {
+            markPermissionAsUnfulfillable(permission);
+            throw new PermissionStartInThePastException(permission.id(), start);
+        }
+    }
```
tformatix, line 450: Why can't a permission start at some point in the future? There is already an implementation of a scheduler that triggers such permissions at a later time.

data-needs/src/main/resources/db/migration/data-needs/V2_5__add_aiida_data_need_contexts.sql
```diff
@@ -0,0 +1,9 @@
+--  SPDX-FileCopyrightText: 2026 The EDDIE Developers <eddie.developers@fh-hagenberg.at>
+--  SPDX-License-Identifier: Apache-2.0
+
+CREATE TABLE aiida_data_need_contexts
+(
+    data_need_id varchar(36) NOT NULL,
```
tformatix, line 6:
```suggestion
    data_need_id uuid NOT NULL,
```

data-needs/src/test/resources/test-data-needs-management-api.sql
```diff
@@ -92,6 +92,13 @@ CREATE TABLE IF NOT EXISTS data_needs.aiida_data_need_permission_commands
     PRIMARY KEY (data_need_id, permission_command)
 );
 
+CREATE TABLE IF NOT EXISTS data_needs.aiida_data_need_contexts
+(
+    data_need_id varchar(36) NOT NULL,
```
tformatix, line 97:
```suggestion
    data_need_id uuid NOT NULL,
```

tformatix (commented):
some tiny remarks - overall very great job! 🥳

tformatix (approved):

data-needs/src/main/resources/db/migration/data-needs/V2_5__add_aiida_data_need_contexts.sql
re1: Used varchar to match existing migrations. There seems to be some reason we do this since there is a migration `V1_8__use_varchar_for_aiida_data_need_id.sql` that specifically replaces existing uuid types with varchar. Not sure what our reason was. This might originate from before we committed to Postgres.

aiida/src/main/java/energy/eddie/aiida/services/PermissionService.java
re1: I looked into this, and we specifically [decided](https://teams.cloud.microsoft/l/message/19:03013ff2-3885-44ae-a428-74a5fc8be855_0a33a163-69fe-409d-b5ba-059c5c6ea7c5@unq.gbl.spaces/1783675479514?context=%7B%22contextType%22%3A%22chat%22%7D) to fail permission requests starting in the past, which was implemented [here](https://github.com/eddie-energy/eddie/commit/8b7c383f2f89118581cdf09c2029f74f95ac90df). However, I agree that this limit seems overly strict and that we might want to reconsider. Maybe towards only failing if the end date is in the past. Will take this into the next refinement.

PR #2548, opened 2026-07-10
GH-2397 Add connection limit API

### LINKED ISSUE #2397: AIIDA: FCA 2.3 - REST API to access current and future connection limits for AIIDA account and permission
**Author**: @georghartner (updated by @re1 based on refinements)

**Priority:** P5

**As an** external application
**I want** to retrieve connection limits (e.g. FCA) via a REST API
**So that** I can obtain and react to current and future connection limits

## Description
Expose a lightweight REST service that provides connection limit values for inbound permissions providing connection limits.

### Current interval
`GET /connection_limits/{permission_id}/now`

Returns the current interval limits for the given permission.

### Future interval
`GET /connection_limits/{permission_id}/plus/{interval_number}`

Returns limits for the interval {interval_number} × 15 minutes ahead.

## Acceptance Criteria
- [x] Response includes:
  - [x] meterId
  - [x] intervalStart (ZonedDateTime[local])
  - [x] minLimit (KW)
  - [x] maxLimit (KW)
- [x] Default interval length: 15 minutes
- [x] API latency < 200 ms
- [x] API secured via OAuth / Keycloak
- [x] API documented via OpenAPI.

re1:
Implements #2397 with the following features:
- Filter by permission and meter ID
- Filter by from to time frame (timestamp or duration)
- Calculate overlaps into one effective timeline per permission and meter

Two revision mechanisms are added that comply with provided best practices for CIM documents:
- If a document ID is reused and the revision number updated, **replace** all entries from that document ID
- If limits exist for a meter and time frame, **update** it if the document creation date time is newer

A filter was chosen instead of a path variable for the permission ID as it simplifies implementation and the connection limit does not always directly map to a set of connection limits.

Both from and to values default to "now" so a query without time frame parameters yields only the currently applicable limits.

As [sweep line algorithm](https://en.wikipedia.org/wiki/Sweep_line_algorithm) was implemented for overlay calculations with some time series specific adaptations.

While this PR includes the groundwork and calculation for default limits, managing these defaults will be implemented in #2588.

I do have some validity checks on the market document, but they are mostly defensive and ensuring correctness is actually out of my scope. I would like to defer this to #2594.

## Assumptions

- Default to current limit if no time frame is provided.
	- Listing _all_ limits ever does not seem useful.
- Migration does not create limits for existing inbound records.

## Screenshots

image


aiida/src/main/java/energy/eddie/aiida/models/connectionlimit/ConnectionLimit.java
```diff
@@ -0,0 +1,112 @@
...
+@IdClass(ConnectionLimit.ConnectionLimitKey.class)
+@Table(name = "connection_limit")
+public class ConnectionLimit {
+
+    @Id
+    @Column(name = "permission_id", nullable = false, updatable = false)
+    private UUID permissionId;
+
+    @Id
+    @Column(name = "meter_id", nullable = false, updatable = false)
```
stefanpenzinger, line 26: Is there a reason to not make it nullable and force empty strings in the Service and Dto?
Just curious 🙂

re1: Has been a while, but I think that during implementation I found handling nullability harder than enforcing an empty string if null :thinking:  Will reconsider when I do a round of refactoring :+1:

re1: There was indeed a reason for this. It was so I can use it as a primary key for simple upserts.

stefanpenzinger (approved):
Well done 👍

PR #2564, opened 2026-07-14
GH-2437: Make it possible to provide Inbound data as MQTT

### LINKED ISSUE #2437: AIIDA: FCA 2.7 - Make it possible to provide Inbound data as MQTT (ex-post configurable and as both MQTT Server and/or MQTT client)
As a s**ystem integrator**
I want to flexibly configure how inbound data is provided (ex-post) using either HTTP or MQTT mechanisms
So that I can integrate AIIDA with different systems and infrastructures using the most suitable communication protocol and security model

## Description

AIIDA shall support multiple configurable inbound data provisioning mechanisms, allowing system integrators to select and modify (ex-post) the preferred integration method per inbound permission or data connection.

The following integration options shall be supported:

- HTTP REST with Bearer Token authentication
- HTTP REST with API Token via GET
- MQTT as Client (AIIDA subscribes to external broker)
- MQTT as Server (AIIDA provides broker access)

The configuration shall be available via the AIIDA UI and/or API (in the inbound permission panel) and can be changed at any time without requiring redeployment.

For MQTT-based integrations, all communication must be secure (TLS) and support credential-based authentication.

## Acceptance Criteria
1. General Configuration
A system integrator can select the inbound integration type from:
HTTP_REST_BEARER
HTTP_REST_API_TOKEN_GET
MQTT_CLIENT
MQTT_SERVER
Configuration is possible ex-post (after initial setup).
Changes take effect without system restart (or within a short propagation time).
Each inbound permission / data connection can have its own integration type.

2. HTTP REST – Bearer TokenAIIDA exposes an HTTP endpoint.
Authentication via Authorization: Bearer <token>.
Token can be generated and managed within AIIDA.
Requests are validated and mapped to the corresponding inbound permission.

3. HTTP REST – API Token (GET)AIIDA exposes an HTTP endpoint accepting requests with an API token via query parameter.
Example: ?apiToken=XYZ
Token is securely generated and linked to the inbound permission.
Access is validated per request.
4. MQTT – Client Mode (AIIDA as Subscriber)
System integrator can configure:
Broker URL
Port
Topic(s)
Username
Password
TLS settings (certificates if required)
AIIDA subscribes to the configured topic(s).
Incoming messages are mapped to the inbound permission.
Connection status and errors are visible in monitoring/logs.
5. MQTT – Server Mode (AIIDA as Broker Provider)
AIIDA reuses its existing MQTT broker capability.
For each inbound permission, AIIDA:
Generates a dedicated username/password
Defines a dedicated topic namespace
System integrator receives connection details:
Broker URL
Port
Username
Password
Topic(s)
Messages published to the topic are ingested by AIIDA.
All connections are secured via TLS.
When the config changes to another channel or the inbound permission is revoked/terminated the MQTT settings should also vanish;
6. Security Requirements
All communication channels must be secured:
HTTPS for REST
TLS for MQTT
Credentials (tokens, usernames, passwords) are:
securely generated
securely stored
revocable / regenerable
Access is restricted per inbound permission.
7. Observability & Operations
Integration status is visible in the UI (connected / disconnected / error).
Logs provide traceability for incoming messages.
Metrics (message count, failures) are available.
8. Backward Compatibility
Existing HTTP-based integrations continue to work unchanged.
Switching integration type does not affect stored data or historical messages.

Flo-Auzi:
GH-2437: Make it possible to provide Inbound data as MQTT

- Add MQTT_SERVER, MQTT_CLIENT, REST_BEARER, REST_API_TOKEN as provisioning modes
- Adapt frontend to provide UI elements to configure and show provisioning modes
- Backend provides endpoints to configure and start different provisioning modes

This is not yet finished as indicated by the draft PR. It provides the structure and functionality for what I thought might be meaningful.
Next steps will include extensive testing, unit tests and documentation


re1 (commented):
Without addressing any individual code segments yet, I am writing together some learnings from our Teams session. I will take a detailed look at the code once we are happy with the overall structure and data flow.

First off, I am making some assumptions for the functionality described by the issue:
- The "system integrator" described in the issue is an AIIDA user wanting to use incoming messages to automate integration of some CU or smart home device.
- "Bearer Token" authentication means API token provided in the header, not user authentication. The HTTP endpoint will be public.
- HTTP provisioning only refers to retrieving the latest message, not streaming all messages via HTTP.
- Provisioning only affects the latest message API endpoint used for automation, not the endpoint providing the UI functionality to download it.

We can confirm our assumptions or bring questions on the issue to the refinement. I generally find it helpful to document assumptions made and questions to refine in the issue or PR for future reference.

For documentation, I suggest you update the [data flow](https://architecture.eddie.energy/aiida/1-running/data-flow.html) and [inbound data source](https://architecture.eddie.energy/aiida/1-running/data-sources/mqtt/inbound/inbound-data-source.html) sections of the AIIDA documentation.

For testing, we can consider an integration or E2E test in addition to unit tests. Since this is quite the long flow our unit tests alone might not ensure the whole flow actually works. Might not be worth the effort though.

I suggest one key structural change in semantically grouping "Provisioning" functionality by name and location. Consider putting all logic related to provisioning inbound data in one package with a separate service and controller building atop the inbound aggregator. This should reduce the mental load needed to understand the feature scope and data flow. The following adaptation to the [data flow diagram](https://architecture.eddie.energy/aiida/1-running/data-flow.html) shows the general approach.

```mermaid
sequenceDiagram
    participant DataSource as Data Source
    participant Adapter as Data Source<br/>Adapter
    participant Aggregator
    participant DB as DB
    participant Provisioner
    participant Scheduler as Permission<br/>Scheduler
    participant Streamer
    actor Client
    actor EP

    DataSource->>Adapter: sends data<br/>(over e.g. MQTT)
    Adapter->>Aggregator: forwards data
    Aggregator->>DB: stores data
    Aggregator->>Provisioner: forwards data
    Scheduler->>Streamer: creates streamers<br/>for permissions
    Aggregator->>Streamer: forwards filtered and processed data
    Provisioner->>Client: streams data over HTTP or MQTT
    Streamer<<->>EP: streams data over MQTT
```

_PlantUML version you can use as reference when updating `data-flow.puml`_

```puml
@startuml

participant DataSource as "Data Source"
participant Adapter as "Data Source\nAdapter"
participant Aggregator
database DB
participant Provisioner
participant Scheduler as "Permission\nScheduler"
participant Streamer
actor Client
actor EP

DataSource -> Adapter : sends data\n(over e.g. MQTT)
Adapter -> Aggregator : forwards data
Aggregator -> DB : stores data
Aggregator -> Provisioner : forwards data
Scheduler -> Streamer : creates streamers\nfor permissions
Aggregator -> Streamer : forwards filtered and processed data
Provisioner -> Client : streams data over HTTP or MQTT
Streamer <-> EP : streams data over MQTT

@enduml
```

aiida/src/main/java/energy/eddie/aiida/services/record/LatestRecordService.java
re1, line 1: Formatting-only change seems accidental. Would not block for this.

aiida/src/main/java/energy/eddie/aiida/models/datasource/mqtt/inbound/InboundDataSource.java
```diff
@@ -28,6 +32,10 @@
 public class InboundDataSource extends MqttDataSource {
     protected static final String TABLE_NAME = "data_source_mqtt_inbound";
 
+    @Transient
+    @JsonIgnore
+    protected String serverModeTopic;
```
re1: Instead of having a transient field with a setter, you can inline this field into the `establishServerModeConnection` reducing the need for the setter and therefore complexity.

```java
    @Transactional
    public ProvisioningConnectionDto establishServerModeConnection(
            MqttConfiguration mqttConfig,
            BCryptPasswordEncoder encoder,
            String plaintextPassword
    ) {
        changeInboundProvisioningType(InboundProvisioningType.MQTT_SERVER);
        var username = UUID.randomUUID().toString();
        var password = Objects.requireNonNull(encoder.encode(plaintextPassword));
        var topic = TOPIC_PREFIX + id + "/inboundData";

        return inboundProvisioningConfig.establishServerModeConnection(
                mqttConfig,
                username,
                password,
                plaintextPassword,
                topic
        );
    }
```

aiida/src/main/java/energy/eddie/aiida/models/datasource/mqtt/inbound/InboundDataSource.java
```diff
@@ -38,26 +46,73 @@ public class InboundDataSource extends MqttDataSource {
...
+            referencedColumnName = "permission_id",
+            insertable = false,
+            updatable = false
+    )
     @JsonIgnore
     private MqttStreamingConfig config;
 
+    @Embedded
+    @JsonProperty("provisioningConfig")
+    private InboundProvisioningConfig inboundProvisioningConfig;
```
re1: I recommend not embedding this field but making it a separate table `data_source_mqtt_inbound_provisioning`. This allows you to
- make this field nullable in case no configuration is needed (REST)
- require the fields inside the config as not nullable (MQTT) to remove some logic (`set`, `clear`) and possibly null checks.

Ping me if you need help with applying this change or don't know where to start 🙂

Since the config only holds information for MQTT provisioning, I would rename it to `MqttProvisioningConfig` or `InboundProvisioningMqttConfig` (if you want to keep the `InboundProvisioning` compound as prefix) which also works well with anticipating other provisioning strategies in the future.

aiida/src/main/java/energy/eddie/aiida/provisioning/ProvisioningMqttPublisher.java
```diff
@@ -0,0 +1,199 @@
...
+                return;
+            }
+
+            client.publish(topic, message);
+        } catch (MqttException ex) {
+            LOGGER.error("Error while publishing message to {} onto topic {}",
+                         mqttConnection.internalHost(),
+                         topic,
+                         ex);
+        }
```
re1: If can be extracted and you might want to log skipping the publish for debugging.

```suggestion
        if (client == null || !client.isConnected()) {
            LOGGER.debug("Cannot publish message to {} onto topic {} since the client is not connected.",
                         mqttConnection.internalHost(),
                         topic)
            return;
        }
        
        try {       
            client.publish(topic, message);
        } catch (MqttException ex) {
            LOGGER.error("Error while publishing message to {} onto topic {}",
                         mqttConnection.internalHost(),
                         topic,
                         ex);
        }
```

aiida/src/main/java/energy/eddie/aiida/provisioning/ProvisioningMqttPublisher.java
```diff
@@ -0,0 +1,199 @@
...
+            }
+        }
+    }
+
+    /**
+     * Indicates whether the publisher currently has an established MQTT connection.
+     *
+     * @return {@code true} when an MQTT client exists and is connected; otherwise {@code false}.
+     */
+    public boolean isConnected() {
```
re1: Method does not seem to be in use?

aiida/src/main/java/energy/eddie/aiida/web/ProvisioningController.java
re1, line 1: I think this should not be a separate controller but part of the `InboundController`. Following our domain-driven approach, each REST controller ideally manages exactly one resource. There is no "Provisioning" resource. We apply a provisioning type on a specific "Inbound Data Source" resource. While not super clean, the `InboundController` is the closest thing we have to manage an inbound data source. Not sure if this overcomplicates things, but we _could_ see the provisioning configuration as a resource, but this is currently not reflected in the entity. See my comment on the `InboundDataSource`/`InboundProvisioningConfig`.

aiida/src/main/java/energy/eddie/aiida/models/datasource/mqtt/inbound/InboundDataSource.java
```diff
@@ -76,17 +131,118 @@ public Set<AiidaSchema> schemas() {
...
+     */
+    @PostLoad
+    private void restoreTransientState() {
+        if (inboundProvisioningConfig == null) {
+            inboundProvisioningConfig = new InboundProvisioningConfig();
+        }
+        setServerModeTopic();
+    }
+
+    private MqttStreamingConfig streamingConfigOrThrow() {
```
re1: What is the reason for adding this check? Theoretically, the `config` field cannot be null since all public constructors set it and JPA ensures that it is always set when loaded from the database. For me, having it resulted in a series of null points, which I assume is from JPA not being able to lazily load the value from the database.

aiida/ui/src/assets/locales/de.json
```diff
@@ -114,6 +115,32 @@
     "showMqttPassword": "MQTT-Passwort anzeigen",
     "hideMqttPassword": "MQTT-Passwort ausblenden",
     "copyMqttPassword": "MQTT-Passwort kopieren",
+    "mqttProvisioning": {
+      "title": "Eingehende Bereitstellung",
```
re1:
```suggestion
      "title": "Bereitstellung konfigurieren",
```

aiida/ui/src/assets/locales/de.json
```diff
@@ -140,6 +167,7 @@
       "permissionId": "Berechtigungs-ID",
       "downloadLatestMessageButton": "Neueste Nachricht herunterladen",
       "inboundTooltip": "Es gibt zwei Möglichkeiten, diesen API-Schlüssel zu verwenden, um die neuesten eingehenden Daten abzurufen:",
+      "provisioningType": "Bereitstellungstyp",
```
re1:
```suggestion
      "provisioningType": "Bereitstellungsmodus",
```

aiida/ui/src/assets/locales/en.json
```diff
@@ -114,6 +115,32 @@
     "showMqttPassword": "Show MQTT Password",
     "hideMqttPassword": "Hide MQTT Password",
     "copyMqttPassword": "Copy MQTT Password",
+    "mqttProvisioning": {
+      "title": "Inbound Provisioning",
```
re1: Remember that the AIIDA UI is user facing and they expect user facing terminology.


```suggestion
      "title": "Configure Provisioning",
```

aiida/ui/src/components/Modals/UpdateMqttProvisioningConnectionModal.vue
re1, line 1: You could just name it `ProvisioningModal`. There should be no naming conflicts for the "Provisioning" concept.

aiida/ui/src/components/PermissionDetails.vue
```diff
@@ -319,13 +358,20 @@ onClickOutside(target, () => (showToolTip.value = false))
...
+      <div v-if="status === 'Active'" class="permission-actions">
+        <div class="actions-row actions-row--end">
+          <MessageDownloadButton :data="permission" class="action-btn">
+            <EyeIcon /> {{ t('permissions.dropdown.downloadLatestMessageButton') }}
+          </MessageDownloadButton>
+        </div>
+        <div v-if="permission.dataNeed.type === 'inbound-aiida'" class="actions-row">
+          <Button button-style="error" class="action-btn" @click="handleRevoke">
+            <RevokeIcon /> {{ t('revokeButton') }}
+          </Button>
```
re1: The revoke button is not just for inbound permissions but also outbound. This change is why your E2E tests fail.

re1 (commented):
Focused on the backend code and high level code flow for now. Will do another round of details once the flow is fully fledged out.

I also have some high level feedback:

- [x] Not sure if all the Javadoc comments are needed. Most of them I find useful. In some instances I think the naming of a class, field, or method could be improved to remove the need for a Javadoc comment. For example, `ProvisioningConnectionDto` → `MqttProvisioningConnectionDto` would make the fields self-explanatory.
	- I noticed you also added comments to some methods you did not change. Fine with doing so if it helps you understand their usage better.
- [x] Suggested this previously, but I would like you to update the [data flow](https://architecture.eddie.energy/aiida/1-running/data-flow.html) and [inbound data source](https://architecture.eddie.energy/aiida/1-running/data-sources/mqtt/inbound/inbound-data-source.html) sections of the AIIDA documentation. Make sure to separate the concepts of "Streaming" to the EP and "Provisioning" to external clients. Please also inform on the default provisioning mode.
- [ ] Please clarify with Georg _if_ and _why_ the strategy for passing the access token via REST is exclusive to either header (Bearer) or body (API token) and document the question and response in the parent issue.
	- [ ] Additionally, we should consider allowing _multiple_ provisioning modes to possibly integrate different devices with the same permission. Theoretically, both REST strategies could be enabled while having both MQTT strategies disabled until configured. We can also defer this as out of scope.
- [x] Please clarify with Georg if we want to have the provisioning as non-optional. I am not sure if this is intended. As of now, the user _has to_ choose a provisioning mode, defaulting to "REST Bearer". I assume this stems from us currently having the API endpoint enabled and an access token generated by default. I think we should use this issue to fully define our provisioning approach in general.

If you want I can take on the mentioned communication steps to clarify requirements.

aiida/ui/src/components/Modals/UpdateMqttProvisioningConnectionModal.vue
Flo-Auzi: Ok sure, I could do that, though I initially thought it might make sense to follow the other Modal namings, which also verbally describe what they are doing.

aiida/ui/src/components/PermissionDetails.vue
Flo-Auzi: Yes, of course, my bad.

aiida/ui/src/components/Modals/UpdateMqttProvisioningConnectionModal.vue
re1: I see. Think I was more confused by the term "Enter". I think the other modal(s) use "Update". We have a bit of a mix here since Alex and I did not sync much on namings :sweat_smile: Choose whatever you find fits :slightly_smiling_face:

aiida/src/main/java/energy/eddie/aiida/web/ProvisioningController.java
re1: Did not read your controller well. As there are only provisioning-specific endpoints and not record-specific ones, my comment is obsolete / plain wrong :upside_down_face:

aiida/src/main/java/energy/eddie/aiida/models/datasource/mqtt/inbound/InboundDataSource.java
Flo-Auzi: Done

aiida/src/main/java/energy/eddie/aiida/models/datasource/mqtt/inbound/InboundDataSource.java
Flo-Auzi: Initially an integration test failed because of some issue that this embedding did not properly restore its values when restarting the application when a MQTT provisioning config was configured and therefore stored in the DB. This is now omitted anyway, so it should not pose a problem anymore.

aiida/ui/src/assets/locales/de.json
Flo-Auzi: Done

aiida/ui/src/assets/locales/de.json
Flo-Auzi: Translation was duplicated anyway, removed it.

aiida/ui/src/assets/locales/en.json
Flo-Auzi: Yes, thanks, done.

aiida/ui/src/components/Modals/UpdateMqttProvisioningConnectionModal.vue
Flo-Auzi: Renamed it to UpdateMqttProvisioningConnectionModal, since I think this also fits the backend naming of what it actually updates.

aiida/ui/src/assets/locales/de.json
Flo-Auzi: Done

aiida/src/main/java/energy/eddie/aiida/provisioning/ProvisioningMqttPublisher.java
Flo-Auzi: It *was* in use, forgot about it. Removed it now

aiida/src/main/java/energy/eddie/aiida/provisioning/ProvisioningMqttPublisher.java
Flo-Auzi: Done, thanks

aiida/src/main/java/energy/eddie/aiida/models/datasource/mqtt/inbound/InboundDataSource.java
Flo-Auzi: Done, please take a look at it again :)

aiida/src/main/java/energy/eddie/aiida/models/datasource/mqtt/inbound/InboundDataSource.java
Flo-Auzi: Done, I also changed the topic to use the username, not the data source id, since I thought it would be more consistent for it to be linked to the additional user that is created especially for provisioning.

aiida/src/main/java/energy/eddie/aiida/services/record/LatestRecordService.java
Flo-Auzi: Should I remove this then? I see that it does not necessarily belong to this PR, but still the formatting now fits our conventions again.

Flo-Auzi:
Regarding your comment about the Javadocs:
I personally believe that we should add more documentation to the code, especially in time of AI this does not pose a huge problem anymore. I do understand that the code itself should be self-explanatory, yet I think it still helps to have a verbal description of different functionalities - I also think this is beneficial with regard to being open source.

If you think this is too much overhead or leads to the code being less readable or maintainable, I can stop adding javadocs again :D

aiida/src/main/java/energy/eddie/aiida/services/record/LatestRecordService.java
re1: I prefer not to include formatting-only changes in feature PRs as they create noise and might misinform about when and why a file was last changed. The PR/commit will state this change was done to support "providing inbound data as MQTT", which it was surely not. While the formatting change itself is not problematic, I think a habit of committing unrelated formatting changes might be. The next person to actually change something in this file will fix the formatting. Might be a bit picky from my side.

aiida/docs/1-running/data-sources/mqtt/inbound/inbound-data-source.md
```diff
@@ -35,27 +36,37 @@ The following schemas are currently supported for inbound data:
...
+2. **REST Bearer:** Retrieves the latest record through the REST interface with the API key supplied in the `X-API-Key` header.
     ```bash
     curl {URL_TO_AIIDA}/inbound/latest/{PERMISSION_ID} \
       --header "X-API-Key: {API_KEY}"
     ```
-   
+3. **MQTT Client:** Connects AIIDA to an external MQTT broker using the host, username, password, and topic entered in the configuration dialog. AIIDA publishes inbound records to that topic.
+4. **MQTT Server:** Uses the MQTT broker managed by AIIDA. AIIDA generates the host, username, password, and topic required to subscribe to inbound records. Save these credentials when they are shown, as the password cannot be displayed again.
+
+The active provisioning mode and, for MQTT provisioning, its connection details are shown in the permission details. The MQTT password is intentionally omitted.
```
re1, line 66: Comma feels off 🤔

```suggestion
The active provisioning mode, and for MQTT provisioning its connection details, are shown in the permission details. The MQTT password is intentionally omitted.
```

aiida/src/main/java/energy/eddie/aiida/models/datasource/mqtt/inbound/InboundDataSource.java
re1: Looks good. As expected you were able to remove a good amount of logic 👍

aiida/src/main/java/energy/eddie/aiida/models/datasource/mqtt/inbound/InboundDataSource.java
```diff
@@ -38,26 +42,80 @@ public class InboundDataSource extends MqttDataSource {
...
+    @Schema(description = "The type defining how inbound data will be provided.")
+    @JsonProperty
+    private InboundProvisioningType provisioningType;
+
+    @OneToOne(fetch = FetchType.EAGER, optional = false)
+    @JoinColumn(
+            name = "permission_id",
+            table = TABLE_NAME,
+            referencedColumnName = "permission_id",
+            insertable = false,
```
re1, line 60: Should the config be insertable? The constructor sets it to be saved with the data source in `DataSourceService#createInboundDataSource`, so I would assume that it is actually inserted, and the info is not correct 🤔 Do not know how JPA uses that information to be honest.

```suggestion
            insertable = true,
```

aiida/src/main/java/energy/eddie/aiida/models/datasource/mqtt/inbound/InboundDataSource.java
re1: We have the `orThrow` pattern multiple times in the AIIDA code. However, as we found in a recent clean code session, such method might hide an exception that should be propagated upstream. Instead of silently throwing a null pointer exception, they should be handled. 

For this `streamingConfigOrThrow` the config should never be null, which would render the whole method not needed. It should be highly unlikely that the field becomes null by accident.

For `mqttProvisioningConfigOrThrow` the value is actually expected to be null. The methods calling it do not know if the provisioning type is configured, and they do not have to. If a method is called that relies on this config while it is null, I suggest throwing a custom `ProvisioningConfigurationException` (or similar) to be propagated all the way up to the controller with `throws` in the method signatures.

aiida/src/main/java/energy/eddie/aiida/services/ProvisioningService.java
```diff
@@ -0,0 +1,265 @@
...
+    private InboundDataSource inboundDataSource(UUID permissionId)
+            throws PermissionNotFoundException, InvalidDataSourceTypeException {
+        var permission = permissionRepository.findById(permissionId)
+                                             .orElseThrow(() -> new PermissionNotFoundException(permissionId));
+
+        if (!(permission.dataSource() instanceof InboundDataSource dataSource)) {
+            throw new InvalidDataSourceTypeException();
+        }
+
+        return dataSource;
```
re1: Slightly prefer inverting the if here. Personal preference.

```suggestion
        if (permission.dataSource() instanceof InboundDataSource dataSource) {
            return dataSource;
        }
        throw new InvalidDataSourceTypeException();

```

re1:
> Regarding your comment about the Javadocs: I personally believe that we should add more documentation to the code, especially in time of AI this does not pose a huge problem anymore. I do understand that the code itself should be self-explanatory, yet I think it still helps to have a verbal description of different functionalities - I also think this is beneficial with regard to being open source.

Not against them entirely. I just want the comments to be actually helpful and in many cases a good variable name can completely remove the need for a comment. As you will know from our clean code sessions, every comment made comes with the responsibility to maintain that comment in addition to the code it is meant to document. For me, they also create noise, as now in addition to reviewing your code, I also have to review your comments, and compared to an LLM, the amount of context my brain can handle is tiny :smile:

Just because AI _can_ easily generate comments from your code given a bit of nudging it does not mean it _should_.

> it still helps to have a verbal description of different functionalities

This is a good and important reason to add a comment! Some of our domain knowledge is too complex to grasp from code alone and comments can definitely help here.

aiida/docs/1-running/data-sources/mqtt/inbound/inbound-data-source.md
fweingartshofer: Auzis version is IMO the better one, since the commas mark the "for MQTT provisioning" in parenthesis.

Flo-Auzi: Yes I also believe this should be correct; the linguistical term for this is "attribute" if I am not completely mistaken.

aiida/src/main/java/energy/eddie/aiida/models/datasource/mqtt/inbound/InboundDataSource.java
Flo-Auzi: This is needed as there is another field, permission, that already references permission_id and Hibernate would not know which column is the owning one and would try to consider both as writable mappings for the same joined column, which is invalid I believe.

re1: Did not notice, thanks!

aiida/docs/1-running/data-sources/mqtt/inbound/inbound-data-source.md
re1: Both are correct, my suggestion is just the way I would have read it would I have seen it without commas :sweat_smile:

aiida/src/main/resources/db/aiida/migration/V9_6__add_inbound_provisioning_type.sql
```diff
@@ -0,0 +1,15 @@
...
+CREATE TABLE data_source_mqtt_inbound_provisioning
+(
+    id                 bigserial PRIMARY KEY,
+    mqtt_connection_id bigint NOT NULL UNIQUE REFERENCES mqtt_connection (id),
+    mqtt_acl_id        bigint NOT NULL UNIQUE REFERENCES data_source_mqtt_acl (id)
+);
+
+-- Add columns for the current provisioning type and its optional MQTT configuration
+ALTER TABLE data_source_mqtt_inbound
+    ADD COLUMN provisioning_type text NOT NULL DEFAULT 'REST_BEARER',
```
re1, line 13: Assuming that the default is added here so no existing integrations break, since `REST_BEARER` was effectively the default before this PR. You can resolve this if that assumption is correct.

aiida/src/main/java/energy/eddie/aiida/models/datasource/mqtt/inbound/InboundDataSource.java
```diff
@@ -41,26 +46,80 @@ public class InboundDataSource extends MqttDataSource {
...
+            updatable = false
+    )
     @JsonIgnore
     private MqttStreamingConfig config;
```
re1, line 66: I realize that I previously encouraged you to make the streaming config non-optional. However, when testing with a number of existing permissions I ran into multiple null pointer exceptions due to this field not being set. I would assume that in a previous version it was possible to have permissions without a streaming config. The database migration does not check if the `optional=false` is actually enforced by the database. 

There are two paths to take here, either we clean up existing databases to ensure that this field is indeed always set, or we assume in the code that it can be null. For my local scenario, I included the following migration to remove "orphaned" data sources.

```sql
--  SPDX-FileCopyrightText: 2026 The EDDIE Developers <eddie.developers@fh-hagenberg.at>
--  SPDX-License-Identifier: Apache-2.0

-- Infer permission id for inbound data sources where missing
UPDATE data_source_mqtt_inbound inbound
SET permission_id = p.permission_id
FROM permission p
WHERE inbound.permission_id IS NULL
  AND p.data_source_id = inbound.id;

-- Delete data sources of inbound type with no inbound data source row
BEGIN;
CREATE TEMP TABLE orphans ON COMMIT DROP AS
SELECT ds.id
FROM data_source ds
         JOIN data_source_mqtt dm ON dm.id = ds.id
         LEFT JOIN data_source_mqtt_inbound inbound ON inbound.id = ds.id
         LEFT JOIN permission p ON p.data_source_id = ds.id
WHERE ds.type = 'INBOUND'
  AND inbound.id IS NULL
  AND p.permission_id IS NULL;

DELETE
FROM data_source_mqtt m USING orphans o
WHERE m.id = o.id;

DELETE
FROM data_source d USING orphans o
WHERE d.id = o.id;
COMMIT;

-- Prevent the creation of inbound data sources without permissions
ALTER TABLE data_source_mqtt_inbound
    ALTER COLUMN permission_id SET NOT NULL;
```

We could also consider this a bug and create a new ticket specifically targeting database inconsistencies following previous migrations.

aiida/ui/src/components/Modals/UpdateMqttProvisioningConnectionModal.vue
```diff
@@ -0,0 +1,339 @@
...
+
+  host.value = connection?.externalHost ?? ''
+  username.value = connection?.username ?? ''
+  password.value = ''
+  topic.value = ''
+
+  modal.value?.showModal()
+}
+
+const validate = () => {
```
re1: There should be no need for custom validation on the frontend. Especially if there is none on the backend. You can use basic input validation with required, or [pattern](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/pattern#specifying_a_pattern) without the need for any JS.

aiida/src/main/java/energy/eddie/aiida/dtos/provisioning/ProvisioningTypePatchDto.java
re1, line 1: I noticed the controller does not do any input validation on this object and the object does not specify any validity constraints. Ideally, the fields should be marked nullable and checked for blank before being passed to the provisioning type handler. Even neater, you could have a separate patch DTO for the MQTT client mode that includes input validation annotations like `@NotBlank`. You could define separate endpoints for this (which would probably conflict with REST principles) or use [Polymorphic Deserialization](https://bwgjoseph.com/polymorphic-dto-using-java-record-with-jackson).

aiida/ui/src/components/Modals/UpdateMqttProvisioningConnectionModal.vue
```diff
@@ -0,0 +1,339 @@
...
+  clearSensitiveState()
+  modal.value?.close()
+}
+
+const showModal = async (targetPermission: AiidaPermission) => {
+  clearSensitiveState()
+  permission.value = targetPermission
+  error.value = ''
+  provisioningType.value = (targetPermission.dataSource?.provisioningType ??
+    'REST_BEARER') as ProvisioningType
```
re1: Cast should not be needed.

```suggestion
  provisioningType.value = targetPermission.dataSource?.provisioningType ?? 'REST_BEARER'
```

re1 (commented):
Following our brief discussion I assume the remaining comments have been kept intentionally following critical consideration.

The PR is already pretty solid. Unfortunately, a PR this big takes considerable brain power to review 😅 I would also like @GoodVibezOnly to take a look at the Vue part 🙂

aiida/src/main/java/energy/eddie/aiida/web/LatestRecordController.java
```diff
@@ -93,8 +94,9 @@ public LatestOutboundPermissionRecordDto latestOutboundPermissionRecord(@PathVar
     @GetMapping(value = "permission/{id}/inbound/latest")
     public LatestInboundPermissionRecordDto latestInboundPermissionRecord(
             @PathVariable("id") UUID permissionId
-    ) throws PermissionNotFoundException, InvalidDataSourceTypeException, InboundRecordNotFoundException,
-             UnsupportedInboundRecordTransformationException, InvalidInboundPermissionException {
+    ) throws PermissionNotFoundException, InvalidDataSourceTypeException,
+             InboundRecordNotFoundException, UnsupportedInboundRecordTransformationException,
+             InvalidInboundPermissionException, ProvisioningTypeNotConfiguredException {
```
GoodVibezOnly: Dont know if you see the sonar issues: Remove the declaration of thrown exception 'energy.eddie.aiida.errors.inbound.ProvisioningTypeNotConfiguredException', as it cannot be thrown from method's body.

aiida/src/test/java/energy/eddie/aiida/models/datasource/mqtt/inbound/InboundDataSourceTest.java
```diff
@@ -12,155 +15,286 @@
...
-        var permission = mock(Permission.class);
-        when(permission.userId()).thenReturn(USER_ID);
-        when(permission.id()).thenReturn(permissionId);
-        when(permission.dataNeed()).thenReturn(dataNeed);
-        when(permission.mqttStreamingConfig()).thenReturn(mock(MqttStreamingConfig.class));
+        var owningPermission = mock(Permission.class);
+        when(owningPermission.userId()).thenReturn(USER_ID);
+        when(owningPermission.id()).thenReturn(permissionId);
+        when(owningPermission.dataNeed()).thenReturn(dataNeed);
+        when(owningPermission.mqttStreamingConfig()).thenReturn(mock(MqttStreamingConfig.class));
```
GoodVibezOnly: SONAR: Extract this mock creation to a local variable.

aiida/src/test/java/energy/eddie/aiida/services/record/LatestRecordServiceTest.java
```diff
@@ -237,8 +238,9 @@ void latestOutboundPermissionRecord_shouldHandleEmptyMessages() throws LatestPer
 
     @Test
     void latestInboundPermissionRecord_shouldReturnLatestRecord_whenFound()
-            throws PermissionNotFoundException, InvalidDataSourceTypeException, InboundRecordNotFoundException,
-                   UnsupportedInboundRecordTransformationException, InvalidInboundPermissionException {
+            throws PermissionNotFoundException, InvalidDataSourceTypeException,
+                   InboundRecordNotFoundException, UnsupportedInboundRecordTransformationException,
+                   InvalidInboundPermissionException, ProvisioningTypeNotConfiguredException {
```
GoodVibezOnly: SONAR: Remove the declaration of thrown exception 'energy.eddie.aiida.errors.inbound.ProvisioningTypeNotConfiguredException', as it cannot be thrown from method's body.

aiida/src/test/java/energy/eddie/aiida/services/record/LatestRecordServiceTest.java
```diff
@@ -261,8 +263,9 @@ void latestInboundPermissionRecord_shouldReturnLatestRecord_whenFound()
 
     @Test
     void latestInboundPermissionRecord_shouldPropagatePermissionNotFoundException()
-            throws PermissionNotFoundException, InvalidDataSourceTypeException, InboundRecordNotFoundException,
-                   UnsupportedInboundRecordTransformationException, InvalidInboundPermissionException {
+            throws PermissionNotFoundException, InvalidDataSourceTypeException,
+                   InboundRecordNotFoundException, UnsupportedInboundRecordTransformationException,
+                   InvalidInboundPermissionException, ProvisioningTypeNotConfiguredException {
```
GoodVibezOnly: SONAR: Remove the declaration of thrown exception 'energy.eddie.aiida.errors.inbound.ProvisioningTypeNotConfiguredException', as it cannot be thrown from method's body.

aiida/src/test/java/energy/eddie/aiida/services/record/LatestRecordServiceTest.java
```diff
@@ -275,8 +278,9 @@ void latestInboundPermissionRecord_shouldPropagatePermissionNotFoundException()
 
     @Test
     void latestInboundPermissionRecord_shouldPropagateInboundRecordNotFoundException()
-            throws PermissionNotFoundException, InvalidDataSourceTypeException, InboundRecordNotFoundException,
-                   UnsupportedInboundRecordTransformationException, InvalidInboundPermissionException {
+            throws PermissionNotFoundException, InvalidDataSourceTypeException,
+                   InboundRecordNotFoundException, UnsupportedInboundRecordTransformationException,
+                   InvalidInboundPermissionException, ProvisioningTypeNotConfiguredException {
```
GoodVibezOnly: SONAR: Remove the declaration of thrown exception 'energy.eddie.aiida.errors.inbound.ProvisioningTypeNotConfiguredException', as it cannot be thrown from method's body.

aiida/src/test/java/energy/eddie/aiida/services/record/LatestRecordServiceTest.java
```diff
@@ -289,8 +293,9 @@ void latestInboundPermissionRecord_shouldPropagateInboundRecordNotFoundException
 
     @Test
     void latestInboundPermissionRecord_shouldPropagateInvalidInboundPermissionException()
-            throws PermissionNotFoundException, InvalidDataSourceTypeException, InboundRecordNotFoundException,
-                   UnsupportedInboundRecordTransformationException, InvalidInboundPermissionException {
+            throws PermissionNotFoundException, InvalidDataSourceTypeException,
+                   InboundRecordNotFoundException, UnsupportedInboundRecordTransformationException,
+                   InvalidInboundPermissionException, ProvisioningTypeNotConfiguredException {
```
GoodVibezOnly: Remove the declaration of thrown exception 'energy.eddie.aiida.errors.inbound.ProvisioningTypeNotConfiguredException', as it cannot be thrown from method's body.

aiida/ui/src/components/Modals/UpdateMqttProvisioningConnectionModal.vue
```diff
@@ -0,0 +1,339 @@
...
+
+  if (!inboundProvisioningTypes.value.provisioningTypes.length) {
+    await fetchInboundProvisioningTypes()
+  }
+
+  const connection = targetPermission.dataSource?.mqttProvisioningConfig?.connection
+
+  host.value = connection?.externalHost ?? ''
+  username.value = connection?.username ?? ''
+  password.value = ''
```
GoodVibezOnly: doesnt clearSensitiveState already reset password?

aiida/ui/src/components/Modals/UpdateMqttProvisioningConnectionModal.vue
```diff
@@ -0,0 +1,339 @@
...
+  ].join('\n')
+})
+
+const clearSensitiveState = () => {
+  provisioningResult.value = undefined
+  password.value = ''
+}
+
+const closeModal = () => {
+  clearSensitiveState()
```
GoodVibezOnly: you call this both on open and close. is that not redundant?

aiida/ui/src/components/Modals/UpdateMqttProvisioningConnectionModal.vue
```diff
@@ -0,0 +1,339 @@
...
+  }
+
+  loading.value = true
+  error.value = ''
+
+  try {
+    const result = await patchInboundProvisioning(permission.value.permissionId, {
+      permissionId: permission.value.permissionId,
+      type: provisioningType.value,
+      host: requiresMqttClientDetails.value ? host.value : '',
```
GoodVibezOnly: do you need the conditional here for host, username, password, topic?
you could probably just use the values directly as they will be "" if no clientdetails were entered right?

aiida/ui/src/components/Modals/UpdateMqttProvisioningConnectionModal.vue
```diff
@@ -0,0 +1,339 @@
...
+import { useI18n } from 'vue-i18n'
+import {
+  fetchInboundProvisioningTypes,
+  inboundProvisioningTypes,
+} from '@/stores/provisioningTypes.ts'
+
+type ProvisioningType = ProvisioningTypePatchDto['type']
+
+const { t } = useI18n()
+const modal = ref<HTMLDialogElement>()
```
GoodVibezOnly:
```suggestion
const modalRef = useTemplateRef('modal')
```

aiida/ui/src/stores/inboundProvisioningTypes.ts
GoodVibezOnly, line 1: Nitpick: would prefer a different name for provisioningTypes since types reminds me always about typescript types. if they are called provisioningTypes in the backend as well then ignore this

aiida/ui/src/components/Modals/UpdateMqttProvisioningConnectionModal.vue
```diff
@@ -0,0 +1,339 @@
...
+      </div>
+
+      <div class="credentials-actions">
+        <Button @click="closeModal">
+          {{ t('permissions.mqttProvisioning.credentialsSaved') }}
+        </Button>
+      </div>
+    </section>
+
+    <form v-else class="mqtt-form" @submit.prevent="handleSubmit">
```
GoodVibezOnly, line 198: im  a bit tempted to suggest to you to just use FormData instead of 4 refs + v-model but im sure you thought more about this then me 😌

GoodVibezOnly (changes requested):
Only looked at frontend files. some comments:

aiida/ui/src/stores/inboundProvisioningTypes.ts
Flo-Auzi: In the backend they are currently called "InboundProvisioningTypes", I would rename them in the frontend therefore as well. That does not resolve your "types"-issue, but at least they would be identical. Hope that's fine.

aiida/ui/src/components/Modals/UpdateMqttProvisioningConnectionModal.vue
Flo-Auzi: If I am changing the provisioning type from Client to anything else, the client data would be sent to the backend again if I removed the conditionals (I believe). This would not necessarily pose an issue since the backend filters properly, but it would still be misleading imo.

aiida/src/main/java/energy/eddie/aiida/web/LatestRecordController.java
Flo-Auzi: I usually see the sonar issues, yes, some of those you mentioned were flying under the radar on my side though :( 
Sorry and thanks

aiida/src/main/java/energy/eddie/aiida/dtos/provisioning/ProvisioningTypePatchDto.java
Flo-Auzi: I adapted the patch to use your suggestion around polymorphic deserialization.

aiida/ui/src/components/Modals/UpdateMqttProvisioningConnectionModal.vue
Flo-Auzi: Correct.

aiida/ui/src/components/Modals/UpdateMqttProvisioningConnectionModal.vue
Flo-Auzi: yes, removed it in here.

aiida/ui/src/components/Modals/UpdateMqttProvisioningConnectionModal.vue
Flo-Auzi: you're right, removed it for close

aiida/ui/src/components/Modals/UpdateMqttProvisioningConnectionModal.vue
Flo-Auzi: Done.

aiida/ui/src/components/Modals/UpdateMqttProvisioningConnectionModal.vue
Flo-Auzi: I do personally like it as it is, but not sure if that's simply subjective 🤔

aiida/src/main/java/energy/eddie/aiida/models/datasource/mqtt/inbound/InboundDataSource.java
Flo-Auzi: I tried to make sure that legacy data sources still work for now. I believe moving this specific problem to a separate issue would make more sense.

aiida/ui/src/components/Modals/UpdateMqttProvisioningConnectionModal.vue
Flo-Auzi: Yes, removed it, thanks.

aiida/src/test/java/energy/eddie/aiida/services/ProvisioningServiceIntegrationTest.java
```diff
@@ -379,9 +380,9 @@ private Permission createPermission() {
     }
 
     private static InboundAiidaLocalDataNeed inboundDataNeed() {
-        var dataNeed = mock(AiidaDataNeed.class);
+        var dataNeed = mock(InboundAiidaDataNeed.class);
```
fweingartshofer, line 383: discussion(mocking): I might be wrong here, so ignore this comment if this is the case.
In my opinion it should be fine to add an all-args constructor to the InboundAiidaDataNeed class to create test instances & co.
Since mocking a POJO seems like more work and higher maintenance costs than just having a constructor to build the POJO.

Flo-Auzi: Solely for this test, I would personally not add an all args constructor. Should the need arise for the InboundAiidaLocalDataNeed to be used across multiple tests, I would totally reconsider, but as of now I think it's not needed.

aiida/src/main/java/energy/eddie/aiida/models/datasource/mqtt/inbound/InboundDataSource.java
```diff
@@ -99,7 +232,28 @@ protected void postPersist() {
...
     protected void createAccessControlEntry() {
-        this.accessControlEntry = new MqttAccessControlEntry(config.username().toString(), config.dataTopic());
+        var streamingConfig = streamingConfig();
+        this.accessControlEntry = new MqttAccessControlEntry(
+                streamingConfig.username().toString(),
+                streamingConfig.dataTopic()
+        );
+    }
+
+    private MqttStreamingConfig streamingConfig() {
```
re1: The method signature does not include the `throws MissingMqttStreamingConfigException` which, as described in my previous `requireNonNull`/`orThrow` comment can, lead to unexpected behavior. I checked all its calls and can state that from the code flow there can is, in this implementation, no scenario in which it should become a problem unless we made an implementation error. However, from a maintainability standpoint it is already problematic that I had to go through the code flow to get to this conclusion.

Ideally, the config would not be nullable. As mentioned earlier there are still some problems with inconsistent database migrations related to this. The main thing that makes it nullable from the permission side is the state before fetched details, since at this point we do not have the permission details. Again following the code flow we _kinda_ know that it cannot be null, but again I had to follow the code flow to see that.

In short, instead of relying on everyone working with this having understood the code flow, I would add the `throws` to the signature even though it will not surface without us making an implementation error.

aiida/ui/src/stores/inboundProvisioningTypes.ts
re1: Maybe rename the file as well.

re1 (commented):
Good work so far. We are almost there. Just two smallish comments 🙂

aiida/src/main/java/energy/eddie/aiida/models/datasource/mqtt/inbound/InboundDataSource.java
Flo-Auzi: I propagated the exception outward as well, since I thought it would not make sense to simply add the exception to the private method signature only.

re1 (approved):
Fine with me now. We can get this merged 🎉

aiida/ui/src/components/Modals/UpdateMqttProvisioningConnectionModal.vue
re1: We can keep it in my opinion. 4 refs is still small enough here.

GoodVibezOnly (approved):
ill allow it 🧐 

good job 👍 this must have been a huge amount of work 💯

PR #2567, opened 2026-07-28
refactor: Split data needs module into multiple smaller modules

fweingartshofer:
This prevents circular dependencies, previously the data need module imported the api module, therefore, it was not possible to use data needs in the api module. With the split modules, it is possible to use data needs in the api module


benjaminellmer (approved):
Looks Good to Me

PR #2570, opened 2026-07-29
fix(eda-region-connector): Fix ponton reconnect issues

fweingartshofer:


tformatix (dismissed):
lgtm :)

re1 (commented):
"Retry Strategy" seems off since it does not seem to follow the strategy pattern and naming it after "the strategy used to infer if a retry is possible" seems a bit farfetched to me. I would therefore assume the separate class for a retry strategy is motivated by us possibly using a more complex retry logic in the future? Not sure if it is worth separating now if it only holds the relatively simple check if the messenger is healthy. If we do not plan on extending, I personally would not separate it.

fweingartshofer:
> "Retry Strategy" seems off since it does not seem to follow the strategy pattern and naming it after "the strategy used to infer if a retry is possible" seems a bit farfetched to me. I would therefore assume the separate class for a retry strategy is motivated by us possibly using a more complex retry logic in the future? Not sure if it is worth separating now if it only holds the relatively simple check if the messenger is healthy. If we do not plan on extending, I personally would not separate it.

I can rename it, no big deal, but I prefer to keep it separate to test it, since it is not completely trivial.

re1:
> I can rename it, no big deal, but I prefer to keep it separate to test it, since it is not completely trivial.

Why not have `isRetryable` protected or test it via the public `PontonMessengerConnectionImpl#sendMessage` which throwing logic is not tested at the moment. The `sendMessage` method itself should probably also be private, or protected for testing, since it is only used inside its own class. 

Can see the extraction being useful if there is potential for it becoming more complex but would not extract it purely for grouping the three tests. You can also do that in a `PontonMessengerConnectionImplTest` class that currently does not exist. Probably personal preference.

re1 (approved):

PR #2574, opened 2026-08-03
GH-2477 simulation region connector extend simulation capabilities for validated historical data

### LINKED ISSUE #2477: Simulation Region Connector: Extend simulation capabilities for validated historical data
## Short Description

Currently, the simulation region connector cannot generate its own validated historical data, but just forwards data given via the scenarios.
This capabilities should be extended, so that the simulation connector can generate validated historical data with plausible load curves.

Base inputs could be:
- Required load curve shape, such as evening peak, flat, etc.
- Season (summer/winter)
- Day type (weekday/weekend)
- Temperature / weather
- Region or customer class

The output should respect:

- Start/end dates
- Required granularity
- Energy type

The generator could be based on:

- a machine learning model, which should be fast, see the [faraday data set](https://huggingface.co/datasets/OpenSynth/cnz-faraday-5.0)
- a probabilistic approach
- etc

## Acceptance Criteria
- [x] Simulation region connector can generate load curves on its own
- [ ] Load curves are based on input parameters
- [ ] And respect start/end, granularity, energy type

fweingartshofer:
This PR adds an additional step to the scenario capabilities of the simulation region connector.
It allows users to generate data based on a normalized load curve profile, which are provided by the simulation region connector.

The seven profiles are based **on** - and not directly taken from, since the numerical values are not provided - the profiles described in this paper: https://doi.org/10.1016/j.apenergy.2024.122683

Use your FH account to access it.
The source graphics are on page 17.

Here the load curve profiles that are actually used in the simulation region connector:

image

Furthermore, this PR adds more documentation and adds a file upload to the frontend, which was vibed.


region-connectors/region-connector-simulation/src/main/java/energy/eddie/regionconnector/simulation/engine/steps/runtime/ValidatedHistoricalDataEmissionStep.java
```diff
@@ -55,4 +55,11 @@ public final boolean equals(Object o) {
 
         return meterReading.equals(that.meterReading);
     }
+
+    @Override
+    public String toString() {
+        return "ValidatedHistoricalDataEmissionStep{" +
+               "meterReading=" + meterReading +
+               '}';
```
re1: Char quotes at last brace? Find easier to read with formatted. I assume you added this for consistency? Would have thought the default `toString` would also result in that string 🤔

```suggestion
        return "ValidatedHistoricalDataEmissionStep{meterReading=%s}".formatted(meterReading);
```

region-connectors/region-connector-simulation/src/main/java/energy/eddie/regionconnector/simulation/engine/steps/loadcurve/StandardProfiles.java
```diff
@@ -0,0 +1,65 @@
...
+            while ((line = reader.readLine()) != null) {
+                String[] values = line.split(",");
+                for (int i = 0; i < values.length; i++) {
+                    columns.get(i).add(new BigDecimal(values[i]));
+                }
+            }
+
+            for (int i = 0; i < headers.length; i++) {
+                profiles.put(headers[i], new Profile(columns.get(i)));
+            }
```
re1: Were you to transpose the input CSV and remove the fixed-width whitespace logic, you could effectively reduce this to one loop (did not test). A transposed CSV without a headline also makes it easier to add or remove lines.

```suggestion
            reader.lines().forEach(line -> {
                var columns = Arrays.asList(line.split(","));
                var header = columns.removeFirst();
                profiles.put(header, new Profile(columns.stream().map(BigDecimal::new).toList()));
            })
```

region-connectors/region-connector-simulation/src/main/java/energy/eddie/regionconnector/simulation/engine/steps/runtime/LoadProfileCurveGenerationStep.java
```diff
@@ -0,0 +1,94 @@
...
+    @Override
+    public SequencedCollection<Step> execute(SimulationContext ctx) throws ExecutionException {
+        if (!(ctx.calculationResult() instanceof ValidatedHistoricalDataDataNeedResult res)) {
+            throw new ExecutionException("Invalid data need %s of type %s".formatted(ctx.dataNeed().id(),
+                                                                                     ctx.dataNeed().type()));
+        }
+        var start = res.energyTimeframe().start().atStartOfDay(ZoneOffset.UTC);
+        var end = res.energyTimeframe().end().plusDays(1).atStartOfDay(ZoneOffset.UTC);
+        var minGranularity = res.granularities().getFirst().duration();
+        var x = Duration.between(start, end).dividedBy(minGranularity);
```
re1: Should probably get a real name.

```suggestion
        var iterations = Duration.between(start, end).dividedBy(minGranularity);
```

region-connectors/region-connector-simulation/src/main/java/energy/eddie/regionconnector/simulation/engine/steps/runtime/LoadProfileCurveGenerationStep.java
```diff
@@ -0,0 +1,94 @@
...
+        var measurements = new ArrayList<Measurement>();
+        for (int i = 0; i < x; i++) {
+            var loopingIndex = Duration.ofMinutes(i * minGranularity.toMinutes());
+            var dow = start.plus(loopingIndex).getDayOfWeek();
+            var profile = Objects.requireNonNull(profiles.get(dow));
```
re1, line 61: While this might not be possible to reach in reality, using `requireNonNull` here might eventually lead to an unchecked `NullPointerException` when refactoring.

region-connectors/region-connector-simulation/src/test/java/energy/eddie/regionconnector/simulation/engine/steps/runtime/LoadProfileCurveGenerationStepTest.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.Optional;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+class LoadProfileCurveGenerationStepTest {
+    @Mock
+    private StandardProfiles standardProfiles;
```
re1, line 35: Why not use an actual instance instead of the mock?

region-connectors/region-connector-simulation/src/test/java/energy/eddie/regionconnector/simulation/engine/steps/loadcurve/ProfileTest.java
```diff
@@ -0,0 +1,114 @@
...
+                new BigDecimal("0.8")
+        );
+        var profile = new Profile(values);
+
+        // When
+        var res = profile.get(Duration.ofMinutes(30), BigDecimal.ONE);
+
+        // Then
+        assertEquals(BigDecimal.valueOf(0.8).setScale(res.scale(), RoundingMode.UNNECESSARY), res);
+    }
```
re1: Not sure if I understand what you test here and why it would not be parametrized 🤔

region-connectors/region-connector-simulation/src/main/web/permission-request-form.js
re1, line 119: Not happy with your JS here. You mentioned generating this. While JS might not be a language of your choice, you can drastically improve this code with mostly language agnostic principles of isolating steps into functions and the use of Errors as exceptions. Did not test this yet.

```js
  handleSubmit(event) {
    event.preventDefault();
    const formData = new FormData(event.target);
    this.submitScenario(formData).catch((error) =>
      this.error(error.message ?? error)
    );
  }

  async postFileScenario(file, metadata) {
    const fileScenarioUrl = `${this.baseUrl}/scenarios/run`;
    let scenario;
    try {
      scenario = JSON.parse(await file.text());
    } catch {
      throw new Error("The uploaded scenario is not valid JSON");
    }

    return await this.postScenario(fileScenarioUrl, { metadata, scenario });
  }

  postSelectedScenario(scenario, metadata) {
    const selectedScenarioUrl = `${this.baseUrl}/scenarios/${scenario.replaceAll("-", " ")}/run`;
    return this.postScenario(selectedScenarioUrl, metadata);
  }

  postScenario(url, body) {
    return fetch(url, {
      method: "POST",
      headers: {
        "Content-Type": "application/json",
      },
      body: JSON.stringify(body),
    });
  }

  async submitScenario(formData) {
    let creationDateTime = formData.get("start-date");
    if (creationDateTime) {
      creationDateTime += "T00:00:00.000Z";
    } else {
      creationDateTime = new Date().toISOString();
    }
    const metadata = {
      connectionId: this.connectionId,
      permissionId: formData.get("permission-id"),
      dataNeedId: this.dataNeedId,
      creationDateTime,
    };
    const file = formData.get("scenario-file");
    const scenario = formData.get("scenario");

    let response;
    if (file?.name) {
      response = await this.postFileScenario(file, metadata);
    } else if (scenario) {
      response = await this.postSelectedScenario(scenario, metadata);
    } else {
      throw new Error("Please select a scenario or upload a scenario file");
    }

    const data = await response.json().catch(() => null);
    if (!response.ok) {
      const error = data.errors?.map((error) => error.message).join("\n");
      throw new Error(
        error ?? "Something went wrong when running the scenario"
      );
    }

    this.notify({
      title: "Executing Scenario!",
      message: "Your scenario is currently being executed.",
      variant: "success",
      duration: 10000,
    });
  }
```

region-connectors/region-connector-simulation/src/main/java/energy/eddie/regionconnector/simulation/engine/steps/runtime/ValidatedHistoricalDataEmissionStep.java
fweingartshofer: This is the autogenerate from intellij toString method, will remove it, since I only used it for debugging.

region-connectors/region-connector-simulation/src/main/java/energy/eddie/regionconnector/simulation/engine/steps/runtime/LoadProfileCurveGenerationStep.java
fweingartshofer: And that's the reason why I added it, since it indicates broken code, that should not exist in the state that it is.

region-connectors/region-connector-simulation/src/main/java/energy/eddie/regionconnector/simulation/engine/steps/loadcurve/StandardProfiles.java
fweingartshofer: This won't work for multiple reasons:

1. Arrays.asList returns fixed size lists, as per the java doc: 
image
2. A CSV file needs to be read top down, where the first line is the header and all consecutive lines contain the values. This implementation assumes there is a complete profile per line.

region-connectors/region-connector-simulation/src/test/java/energy/eddie/regionconnector/simulation/engine/steps/runtime/LoadProfileCurveGenerationStepTest.java
fweingartshofer: Because I don't want to test parsing of the csv and the actual profiles contain actual values, which I would have to calculate beforehand for the test.
Instead I use a static profile which will always return 1.

region-connectors/region-connector-simulation/src/test/java/energy/eddie/regionconnector/simulation/engine/steps/loadcurve/ProfileTest.java
fweingartshofer: Profiles with only a singular item are tested with a granularity of less than an hour.

region-connectors/region-connector-simulation/src/main/java/energy/eddie/regionconnector/simulation/engine/steps/runtime/LoadProfileCurveGenerationStep.java
re1: Not sure if there is an intended way to use it, but I would think requireNonNull is for when you also want to catch the NullPointerException generated from it, which is not done here. Could very well be wrong here. It actually feels like a case for an "assert", but asserts are very much frowned upon in the Java world, and I was told to never use them :thinking:

region-connectors/region-connector-simulation/src/main/java/energy/eddie/regionconnector/simulation/engine/steps/loadcurve/StandardProfiles.java
re1: There surely is a neat implementation that does not use `Arrays.asList`. Wrote the suggestion without trying it.

To my knowledge, there is no agreed on standard for CSV like that. What I find so neat about CSV is that it is exactly what it says it is: values separated by commas.

region-connectors/region-connector-simulation/src/main/java/energy/eddie/regionconnector/simulation/engine/steps/runtime/LoadProfileCurveGenerationStep.java
fweingartshofer: I didn't know that you always had to use a requireNonNull with a corresponding catch statement, could you point me to an article that describes the reasoning?
Because as far as I have seen it used in code bases, it is used as a validation mechanism during start up/wiring, as well as a mechanism to validate that something should never be null at a specific point in a program.
If it is null, it means that the state of the program is incorrect and should be stopped.
In this case, there is validation that the profiles in the scenarios actually exist, if they do not exist, the user will get an error message.
During execution it is assumed that the validation was successful, and the scenario can be executed without a second thought.

**Update**: Previously there was a check if a certain profile exists in line 90 of the file.
I updated the exception to be a checked exception, I think this should suffice.
At some point it is only defensive programming for the sake of it.

fweingartshofer:
Build is fixed in #2578

region-connectors/region-connector-simulation/src/main/java/energy/eddie/regionconnector/simulation/engine/steps/runtime/LoadProfileCurveGenerationStep.java
re1: You do not have to and your IDE will not warn you since the signature does not define a `throws`. These are only my thoughts based on my experiences. There surely are more than enough articles in favor or against either approach. I have not read a single one.

I am also just talking about run-time here. For startup, I see no problem in not catching the exception, since it will not impact anything at run-time. What I do not like is ignoring it at run-time. If you _know_ that "something should never be null at a specific point in a program", that to me sounds like an `assert`. If you are reasonably sure that something cannot happen but want to be ready if it happens anyway, that to me sounds like an `InvalidStateException`.

> If it is null, it means that the state of the program is incorrect and should be stopped. In this case, there is validation that the profiles in the scenarios actually exist, if they do not exist, the user will get an error message.

Setting aside that we both see this as an unreachable statement, the problem is that _should_ you reach it, since this is at run-time, the execution of the scenario will just stop, and it will only show up as a generic NPE in the logs.

fweingartshofer: asserts are removed for production builds, so it cannot be used.

I would much prefer to skip this argument, since this will lead to needlessly defensive coding practices, which are a code smell.

re1: We can leave the line as it is if you also find it not worth the effort of discussing.

I do not reference any specific detail to avoid unnecessary arguing, but I still do not agree with some of your statements and find they partially conflict with your implementation. Maybe because I fail to read them the way you intended, because I failed to communicate my points effectively, or because I simply lack the experience with the language.

region-connectors/region-connector-simulation/src/main/java/energy/eddie/regionconnector/simulation/engine/constraints/LoadProfileCurveStepConstraint.java
```diff
@@ -0,0 +1,32 @@
...
+public class LoadProfileCurveStepConstraint implements ElementConstraint {
+
+    @Override
+    public ConstraintResult violatesConstraint(Model model) {
+        if (!(model instanceof LoadProfileCurveStep step)) {
+            return new ConstraintOk();
+        }
+        var stdProfiles = StandardProfiles.getInstance();
+        var allProfiles = String.join(", ", stdProfiles.allProfiles());
+        var missingProfiles = step.allProfiles()
```
stefanpenzinger:
```suggestion
        var unknownProfiles = step.allProfiles()
```
Should it be called like that, given that they are unknown to the system rather than missing?

stefanpenzinger (dismissed):
Great job!
Just a small comment.

re1 (approved):

Issue #2540, opened 2026-07-02
Data Needs: Allow EP to omit fields in config file that can be empty


re1:
## Short Description
**As** an eligible party **I want** to omit fields in the data needs config file that I do not need **so that** my config file becomes (1) easier to read and (2) more robust against changes to the data need specification.

When configuring a data need from a config file like `data-needs.json`, all fields that not nullable must be present in the config file. Especially for AIIDA, there are fields that can be left empty or are not read at all for specific data needs. Specifically:
- `acknowledgementRequired` → can default to false
- `allowedPermissionCommands` → empty means allow none
- `assets` → #2519
- `dataTags` → #2522
- `contexts` → #2395

Since all these fields can be empty, they should also be optional. This can either be achieved by mapping missing fields to empty values on read or by adapting the data need specifications to make fields that can be empty nullable.

This has become more relevant with the addition of two new fields in the context of FCAs and changes to existing fields becoming lists instead of string values (`allowedPermissionCommands`, `assets`).

`dataTags` could be moved to be outbound only.

The data needs management API may be adjusted to also have the fields as optional. The admin console might need changes as well if the fields become nullable instead of defaulting to empty.

## Acceptance Criteria
- [ ] List-type fields that can be empty are also optional
- [ ] Omitted list-type fields should be functionally equivalent to an empty list (`dataTags: null` ≡ `dataTags: []`)
- [ ] Data needs management API is adapted in case the fields become nullable on entity level
- [ ] Admin console is adapted in case the management API changes

PR #2577, opened 2026-08-07
 refactor(aiida): Move data tags to dedicated interface for outbound data needs

### LINKED ISSUE #2522: AIIDA: Do not filter records by data tag if none were specified in an AIIDA data need
## Short Description

**As** an eligible party **I want** to request any data regardless of its data tag **so that**:
1. I do not have to specify all data tags separately when my application can handle any incoming data tag.
2. I can adapt my application to process additional data tags without the need to create a new data need and request a new permission.

At the moment, all records are filtered to only include requested data tags before being sent. If no data tag was requested (unset or empty list), all data tags should be allowed.

ℹ️ Data tags only apply to [outbound meter data](https://architecture.eddie.energy/aiida/1-running/schemas/schemas.html#outbound-data) and have no effect on inbound schemas.

## Acceptance Criteria
- [x] The data tags field can be omitted from data needs
- [x] The approach is documented in the [AIIDA operation manual](https://architecture.eddie.energy/aiida/1-running/data-need.html)

stefanpenzinger:
This PR documents that data tags are optional and moves data tags to a dedicated outbound data need interface as it is not used for inbound data needs

### DIFF
--- a/aiida/src/main/java/energy/eddie/aiida/models/permission/dataneed/AiidaLocalDataNeedFactory.java
+++ b/aiida/src/main/java/energy/eddie/aiida/models/permission/dataneed/AiidaLocalDataNeedFactory.java
@@ -1,17 +1,20 @@
-// SPDX-FileCopyrightText: 2025 The EDDIE Developers <eddie.developers@fh-hagenberg.at>
+// SPDX-FileCopyrightText: 2025-2026 The EDDIE Developers <eddie.developers@fh-hagenberg.at>
 // SPDX-License-Identifier: Apache-2.0
 
 package energy.eddie.aiida.models.permission.dataneed;
 
 import energy.eddie.dataneeds.needs.aiida.AiidaDataNeed;
 import energy.eddie.dataneeds.needs.aiida.InboundAiidaDataNeed;
+import energy.eddie.dataneeds.needs.aiida.OutboundAiidaDataNeed;
 
 public class AiidaLocalDataNeedFactory {
     private AiidaLocalDataNeedFactory() {}
 
     public static AiidaLocalDataNeed create(AiidaDataNeed dataNeed) {
-        return InboundAiidaDataNeed.DISCRIMINATOR_VALUE.equals(dataNeed.type())
-                ? new InboundAiidaLocalDataNeed(dataNeed)
-                : new OutboundAiidaLocalDataNeed(dataNeed);
+        return switch (dataNeed) {
+            case InboundAiidaDataNeed inbound -> new InboundAiidaLocalDataNeed(inbound);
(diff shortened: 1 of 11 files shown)


fweingartshofer (dismissed):

aiida/src/main/java/energy/eddie/aiida/models/permission/dataneed/AiidaLocalDataNeedFactory.java
```diff
@@ -1,17 +1,18 @@
...
 package energy.eddie.aiida.models.permission.dataneed;
 
 import energy.eddie.dataneeds.needs.aiida.AiidaDataNeed;
 import energy.eddie.dataneeds.needs.aiida.InboundAiidaDataNeed;
+import energy.eddie.dataneeds.needs.aiida.OutboundAiidaDataNeed;
 
 public class AiidaLocalDataNeedFactory {
     private AiidaLocalDataNeedFactory() {}
 
     public static AiidaLocalDataNeed create(AiidaDataNeed dataNeed) {
```
fweingartshofer, line 13: nitpick: This can be replaced by a switch statement and pattern matching instead of the string check.

aiida/src/main/java/energy/eddie/aiida/models/permission/dataneed/AiidaLocalDataNeedFactory.java
```diff
@@ -1,17 +1,18 @@
...
         return InboundAiidaDataNeed.DISCRIMINATOR_VALUE.equals(dataNeed.type())
                 ? new InboundAiidaLocalDataNeed(dataNeed)
-                : new OutboundAiidaLocalDataNeed(dataNeed);
+                : new OutboundAiidaLocalDataNeed((OutboundAiidaDataNeed) dataNeed);
```
re1: Since we are already casting here, maybe we should replace the check of the DISCRIMINATOR_VALUE with a switch statement? 🤔

re1 (commented):
Overall good work 👍

Just one minor comment that we could also ignore.

For the streamer, I think the checks now are already an improvement, but we should consider a permission-type aware refactoring. I want to link #2540 as a related issue, but we can also address it separately.

aiida/src/main/java/energy/eddie/aiida/models/permission/dataneed/AiidaLocalDataNeedFactory.java
re1: Noticed this is a duplicate of Florian's comment which my GitHub did not render correctly :sweat_smile:

re1 (approved):

PR #2580, opened 2026-08-11
feat(aiida): Introduce custom display name for permissions

### LINKED ISSUE #2521: AIIDA: Allow setting a custom display name for a permission
## Short Description
**As** an AIIDA user **I want** to set a custom display name for my permissions **so that** I can choose a name (1) that makes more sense to _me_ than the name of the data need and (2) to distinguish multiple permissions with the same data need without having to look up the permission ID.

## Acceptance Criteria
- [x] Can be set when first adding a permission
- [x] Can be changed for existing permissions
- [x] Defaults to the name of the data need
- [x] Is used when selecting an inbound data source by its permission
- [x] Is used to select permissions in E2E tests

stefanpenzinger:
PR introduces a custom display name for permissions.

| |                                                                                                                                                                                                     
  |---|                                                                                                                                                                                                   
  | <img width="800" alt="20260811-GH2521-01" src="https://github.com/user-attachments/assets/25926773-d8b0-4790-8201-8e2e3ee1c0a0" /> |                                                                  
  | <img width="800" alt="20260811-GH2521-02" src="https://github.com/user-attachments/assets/9cc0bbee-5e70-44d5-b1bd-4395f918e999" /> |                                                                  
  | <img width="800" alt="20260811-GH2521-03" src="https://github.com/user-attachments/assets/d44fdc8e-4efb-4650-b731-a97dea7b6afd" /> |                                                                  
  | <img width="800" alt="20260811-GH2521-04" src="https://github.com/user-attachments/assets/0a56dc90-cfb4-426c-9500-e55be6f9051f" /> |                                                                  
  | <img width="800" alt="20260811-GH2521-05" src="https://github.com/user-attachments/assets/6e360d82-02d8-4e27-bffc-f798d3bb6058" /> |                                                                  
  | <img width="800" alt="20260811-GH2521-06" src="https://github.com/user-attachments/assets/e6237196-5d20-4ed9-8bd8-f54074c09e5e" /> |
| <img width="1013" height="849" alt="20260811-GH2521-07" src="https://github.com/user-attachments/assets/f10a1c82-2994-4a06-ab68-1751080a719a" /> |


aiida/src/main/java/energy/eddie/aiida/models/permission/Permission.java
fweingartshofer: note (code-smell): This setter does more than just setting the data need, it sets the service name, display name, and the inbound message format.
This can lead to unintended side effects.
My suggestion would be to rename it to a more fitting name, so the side effects are clearer.
Can be ignored for now.

aiida/ui/src/components/Modals/UpdatePermissionModal.vue
```diff
@@ -59,12 +63,13 @@ const dataSourceOptions = computed(() => {
...
   if (requestedSchemas.some((requestedSchema) => inboundSchemas.has(requestedSchema))) {
-    for (const { dataSource } of inboundPermissions.value) {
+    for (const inboundPermission of inboundPermissions.value) {
+      const { dataSource } = inboundPermission
       if (
         dataSource && // For TypeScript, even though it should always be set here
         requestedSchemas.some((requestedSchema) => dataSource.schemas?.includes(requestedSchema))
       ) {
-        matches.push({ label: dataSource.name, value: dataSource.id })
+        matches.push({ label: inboundPermission.displayName, value: dataSource.id })
```
re1: You can keep the top-level destructuring

```suggestion
    for (const { dataSource, displayName } of inboundPermissions.value) {
     if (
       dataSource && // For TypeScript, even though it should always be set here
       requestedSchemas.some((requestedSchema) => dataSource.schemas?.includes(requestedSchema))
     ) {
        matches.push({ label: displayName, value: dataSource.id })
```

e2e-tests/src/test/java/energy/eddie/e2etests/aiida/AiidaUiTests.java
```diff
@@ -246,24 +247,25 @@ private void addPermission(String aiidaCode) {
...
-    private String acceptOutboundPermissionRequest(String aiidaCode, String dataSource) {
+    private String acceptOutboundPermissionRequest(String aiidaCode, String dataSourceOrInboundPermissionDisplayName) {
         addPermission(aiidaCode);
 
         var dialog = page.getByRole(AriaRole.DIALOG);
         var id = dialog.locator(":text('Permission ID') + dd").textContent();
 
         dialog.getByRole(AriaRole.LISTBOX).getByText("Select Data Source").click();
-        dialog.getByRole(AriaRole.OPTION).getByText(dataSource).first().click();
+        dialog.getByRole(AriaRole.OPTION).getByText(dataSourceOrInboundPermissionDisplayName).first().click();
```
re1: Do not think there is a need to adjust the name here since an inbound data source labelled with a display name is semantically the same thing as a named outbound data source.

```suggestion
    private String acceptOutboundPermissionRequest(String aiidaCode, String dataSource) {
addPermission(aiidaCode);

var dialog = page.getByRole(AriaRole.DIALOG);
var id = dialog.locator(":text('Permission ID') + dd").textContent();

dialog.getByRole(AriaRole.LISTBOX).getByText("Select Data Source").click();
        dialog.getByRole(AriaRole.OPTION).getByText(dataSource).first().click();
```

aiida/ui/src/components/ConfirmDialog.vue
re1, line 1: What is the reason you had to add this? 🤔

aiida/ui/src/components/PermissionDropdown.vue
re1, line 66: I think you already mentioned this issue (bug?) last week. The permission details still show the data source ID when an inbound permission is selected. A key motivation for this issue was for the user to be presented with the custom name when selecting their inbound permission. This PR is probably a good place to address this.

I simply created an inbound permission from the demo button. The inbound permission name shows up correctly in the dropdown but not the permission details.

Image

re1, line 1: I think we could extract the name editing logic into its own component to reduce the load on the permission dropdown 🤔

re1 (commented):
Played around with this and it works really well 👍

aiida/ui/src/components/ConfirmDialog.vue
stefanpenzinger: In order to reuse the ConfirmDialog, I need to give it a fitting colour. Otherwise, if I want to change the name of the permission, I have to click a red button, which suggests a negative or destructive action.

aiida/src/main/java/energy/eddie/aiida/models/permission/Permission.java
stefanpenzinger: Done :+1:

e2e-tests/src/test/java/energy/eddie/e2etests/aiida/AiidaUiTests.java
stefanpenzinger: Reverted it

aiida/ui/src/components/PermissionDropdown.vue
stefanpenzinger: Done :+1: 
image

aiida/ui/src/components/PermissionDetails.vue
```diff
@@ -113,6 +113,14 @@ const generateStringFromLength = (length: number, char: string) => {
...
 
 onClickOutside(target, () => (showToolTip.value = false))
+
+const dataSourceDisplayName = computed(() => {
+  if (!permission.dataSource) return undefined
+  const inboundPermission = permissions.value.find(
+    (p) => p.dataNeed.type === 'inbound-aiida' && p.dataSource?.id === permission.dataSource?.id,
+  )
+  return inboundPermission?.displayName ?? permission.dataSource.name
+})
```
re1, line 123: What is the searching for? 

```suggestion
const dataSourceDisplayName = computed(() =>
  permission.dataNeed.type !== 'inbound-aiida'
    ? permission.dataSource?.name
    : (permission.displayName ?? permission.dataSource?.name),
)
```

stefanpenzinger: AiidaDataSource has no displayName, so the name of the inbound data source is retrieved from the display name of the inbound permission.
The frontend data model does not have a link/FK from the data source to the owning permission (as it is not exposed by the backend).

**Code sample of yours:** So when we are now trying to forward an inbound permission, the data need type is **outbound**, thus you would always display the **name** (in the case of inbound permissions the ID) of the inbound data source and not the display name of the inbound permission.

The "searching" uses all the loaded permissions to find the actual display name of the inbound permission, which will then be displayed as the data source of the outbound permission forwarding it.

Quite complicated, isn't it :D

re1: Got it. Did not think about the context enough. Probably a candidate for refactoring though :thinking:

stefanpenzinger: yes, could be done in the future

re1 (approved):

PR #2582, opened 2026-08-13
feat(at-eda): improve Ponton WebSocket health and reconnect handling

nathaniel-boisgard:
## What changed

- track adapter-side Ponton WebSocket connection counts and expose them through health checks
- add a configurable watchdog that detects stale or insufficient connections and reconnects safely
- serialize send, reconnect, and shutdown operations to avoid reconnect races
- preserve retry behavior for retryable send failures and prevent reconnects after shutdown
- add focused tests for watchdog scheduling, connection-state evaluation, retries, reconnects, and idempotent shutdown

## Why

The existing health endpoint only showed whether the Ponton Messenger process was reachable. It did not detect stale adapter WebSocket connections, and concurrent send, reconnect, and shutdown paths could race with one another.

This change uses the Ponton Adapter API connection-status callback to monitor the actual inbound, outbound, and archive WebSocket connections and recover them when they become unhealthy.

## Impact

AT EDA connections can recover automatically from stale WebSocket sessions. Existing installations retain compatible defaults: one inbound connection, one outbound connection, no archive connection, a one-minute watchdog interval, and a two-minute initial status timeout.

## Validation

- focused watchdog, connection-state, retry, reconnect, and shutdown tests
- AT EDA Java compilation
- JaCoCo coverage verification
- staged diff whitespace validation

### DIFF
--- a/region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/PontonXPAdapterConfiguration.java
+++ b/region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/PontonXPAdapterConfiguration.java
@@ -3,9 +3,13 @@
 
 package energy.eddie.regionconnector.at.eda.ponton;
 
+import jakarta.validation.constraints.Positive;
+import org.hibernate.validator.constraints.time.DurationMin;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.bind.ConstructorBinding;
+import org.springframework.boot.context.properties.bind.DefaultValue;
 import org.springframework.boot.context.properties.bind.Name;
+import org.springframework.validation.annotation.Validated;
 
 import java.time.Duration;
 
@@ -24,10 +28,10 @@ import java.time.Duration;
  * @param password       Password for the given username.
  * @param inboundConnections Expected amount of inbound WebSocket connections to the Ponton XP Messenger.
  * @param outboundConnections Expected amount of outbound WebSocket connections to the Ponton XP Messenger.
- * @param archiveConnections Expected amount of archive WebSocket connections to the Ponton XP Messenger.
  * @param connectionWatchdogInterval How often the Ponton XP Messenger adapter WebSocket connection is checked.
  * @param connectionStatusTimeout Maximum time after start or reconnect until a connection-status callback must be received.
  */
+@Validated
 @ConfigurationProperties("region-connector.at.eda.ponton.messenger")
 public record PontonXPAdapterConfiguration(
         @Name("adapter.id") String adapterId,
@@ -38,55 +42,16 @@ public record PontonXPAdapterConfiguration(
         @Name("folder") String workFolder,
         @Name("username") String username,
         @Name("password") String password,
-        @Name("inbound-connections") int inboundConnections,
-        @Name("outbound-connections") int outboundConnections,
-        @Name("archive-connections") int archiveConnections,
-        @Name("connection-watchdog.interval") Duration connectionWatchdogInterval,
-        @Name("connection-watchdog.status-timeout") Duration connectionStatusTimeout
+        @Name("inbound-connections") @Positive @DefaultValue("1") int inboundConnections,
+        @Name("outbound-connections") @Positive @DefaultValue("1") int outboundConnections,
+        @Name("connection-watchdog.interval")
+        @DurationMin(nanos = 1)
+        @DefaultValue("1m") Duration connectionWatchdogInterval,
+        @Name("connection-watchdog.status-timeout")
+        @DurationMin(nanos = 1)
+        @DefaultValue("2m") Duration connectionStatusTimeout
 ) {
-    private static final int DEFAULT_INBOUND_CONNECTIONS = 1;
-    private static final int DEFAULT_OUTBOUND_CONNECTIONS = 1;
-    private static final int DEFAULT_ARCHIVE_CONNECTIONS = 0;
-    private static final Duration DEFAULT_CONNECTION_WATCHDOG_INTERVAL = Duration.ofMinutes(1);
-    private static final Duration DEFAULT_CONNECTION_STATUS_TIMEOUT = Duration.ofMinutes(2);
-
     @ConstructorBinding
     public PontonXPAdapterConfiguration {
-        inboundConnections = inboundConnections <= 0 ? DEFAULT_INBOUND_CONNECTIONS : inboundConnections;
-        outboundConnections = outboundConnections <= 0 ? DEFAULT_OUTBOUND_CONNECTIONS : outboundConnections;
-        archiveConnections = Math.max(archiveConnections, DEFAULT_ARCHIVE_CONNECTIONS);
-        connectionWatchdogInterval = connectionWatchdogInterval == null
-                ? DEFAULT_CONNECTION_WATCHDOG_INTERVAL
-                : connectionWatchdogInterval;
-        connectionStatusTimeout = connectionStatusTimeout == null
-                ? DEFAULT_CONNECTION_STATUS_TIMEOUT
-                : connectionStatusTimeout;
-    }
-
-    public PontonXPAdapterConfiguration(
-            String adapterId,
-            String adapterVersion,
-            String hostname,
-            int port,
-            String apiEndpoint,
-            String workFolder,
-            String username,
-            String password
-    ) {
-        this(
-                adapterId,
-                adapterVersion,
-                hostname,
-                port,
-                apiEndpoint,
-                workFolder,
-                username,
-                password,
-                DEFAULT_INBOUND_CONNECTIONS,
-                DEFAULT_OUTBOUND_CONNECTIONS,
-                DEFAULT_ARCHIVE_CONNECTIONS,
-                DEFAULT_CONNECTION_WATCHDOG_INTERVAL,
-                DEFAULT_CONNECTION_STATUS_TIMEOUT
-        );
     }
 }
--- a/region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/PontonXPAdapterConfiguration.java
+++ b/region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/PontonXPAdapterConfiguration.java
@@ -4,7 +4,10 @@
 package energy.eddie.regionconnector.at.eda.ponton;
 
 import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.bind.ConstructorBinding;
 import org.springframework.boot.context.properties.bind.Name;
+
+import java.time.Duration;
 
 /**
  * This record defines all information needed for a PontonXPAdapter to establish a connection to a Ponton XP Messenger.
@@ -19,6 +22,11 @@ import org.springframework.boot.context.properties.bind.Name;
  * @param username       Username for the Ponton XP Messenger to use REST API endpoints that require authentication.
  *                       Needs to be a user without 2FA activated.
  * @param password       Password for the given username.
+ * @param inboundConnections Expected amount of inbound WebSocket connections to the Ponton XP Messenger.
+ * @param outboundConnections Expected amount of outbound WebSocket connections to the Ponton XP Messenger.
+ * @param archiveConnections Expected amount of archive WebSocket connections to the Ponton XP Messenger.
+ * @param connectionWatchdogInterval How often the Ponton XP Messenger adapter WebSocket connection is checked.
+ * @param connectionStatusTimeout Maximum time after start or reconnect until a connection-status callback must be received.
  */
 @ConfigurationProperties("region-connector.at.eda.ponton.messenger")
 public record PontonXPAdapterConfiguration(
@@ -29,5 +37,56 @@ public record PontonXPAdapterConfiguration(
         @Name("api.endpoint") String apiEndpoint,
         @Name("folder") String workFolder,
         @Name("username") String username,
-        @Name("password") String password
-) {}
+        @Name("password") String password,
+        @Name("inbound-connections") int inboundConnections,
+        @Name("outbound-connections") int outboundConnections,
+        @Name("archive-connections") int archiveConnections,
+        @Name("connection-watchdog.interval") Duration connectionWatchdogInterval,
+        @Name("connection-watchdog.status-timeout") Duration connectionStatusTimeout
+) {
+    private static final int DEFAULT_INBOUND_CONNECTIONS = 1;
+    private static final int DEFAULT_OUTBOUND_CONNECTIONS = 1;
+    private static final int DEFAULT_ARCHIVE_CONNECTIONS = 0;
+    private static final Duration DEFAULT_CONNECTION_WATCHDOG_INTERVAL = Duration.ofMinutes(1);
+    private static final Duration DEFAULT_CONNECTION_STATUS_TIMEOUT = Duration.ofMinutes(2);
+
+    @ConstructorBinding
+    public PontonXPAdapterConfiguration {
+        inboundConnections = inboundConnections <= 0 ? DEFAULT_INBOUND_CONNECTIONS : inboundConnections;
+        outboundConnections = outboundConnections <= 0 ? DEFAULT_OUTBOUND_CONNECTIONS : outboundConnections;
+        archiveConnections = Math.max(archiveConnections, DEFAULT_ARCHIVE_CONNECTIONS);
+        connectionWatchdogInterval = connectionWatchdogInterval == null
+                ? DEFAULT_CONNECTION_WATCHDOG_INTERVAL
+                : connectionWatchdogInterval;
+        connectionStatusTimeout = connectionStatusTimeout == null
+                ? DEFAULT_CONNECTION_STATUS_TIMEOUT
+                : connectionStatusTimeout;
+    }
+
+    public PontonXPAdapterConfiguration(
+            String adapterId,
+            String adapterVersion,
+            String hostname,
+            int port,
+            String apiEndpoint,
+            String workFolder,
+            String username,
+            String password
+    ) {
+        this(
+                adapterId,
+                adapterVersion,
+                hostname,
+                port,
+                apiEndpoint,
+                workFolder,
+                username,
+                password,
+                DEFAULT_INBOUND_CONNECTIONS,
+                DEFAULT_OUTBOUND_CONNECTIONS,
+                DEFAULT_ARCHIVE_CONNECTIONS,
+                DEFAULT_CONNECTION_WATCHDOG_INTERVAL,
+                DEFAULT_CONNECTION_STATUS_TIMEOUT
+        );
+    }
+}
--- a/region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/PontonXPAdapterConfiguration.java
+++ b/region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/PontonXPAdapterConfiguration.java
@@ -6,7 +6,6 @@ package energy.eddie.regionconnector.at.eda.ponton;
 import jakarta.validation.constraints.Positive;
 import org.hibernate.validator.constraints.time.DurationMin;
 import org.springframework.boot.context.properties.ConfigurationProperties;
-import org.springframework.boot.context.properties.bind.ConstructorBinding;
 import org.springframework.boot.context.properties.bind.DefaultValue;
 import org.springframework.boot.context.properties.bind.Name;
 import org.springframework.validation.annotation.Validated;
@@ -51,7 +50,4 @@ public record PontonXPAdapterConfiguration(
         @DurationMin(nanos = 1)
         @DefaultValue("2m") Duration connectionStatusTimeout
 ) {
-    @ConstructorBinding
-    public PontonXPAdapterConfiguration {
-    }
 }
--- a/region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImpl.java
+++ b/region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImpl.java
@@ -32,7 +32,10 @@ import org.springframework.oxm.XmlMappingException;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.time.Duration;
+import java.time.Instant;
 import java.time.ZonedDateTime;
+import java.util.HashMap;
 import java.util.concurrent.locks.ReentrantLock;
 
 public class PontonMessengerConnectionImpl implements AutoCloseable, PontonMessengerConnection {
@@ -40,10 +43,13 @@ public class PontonMessengerConnectionImpl implements AutoCloseable, PontonMesse
     private final InboundMessageFactoryCollection inboundMessageFactoryCollection;
     private final OutboundMessageFactoryCollection outboundMessageFactoryCollection;
     private final MessengerConnection.MessengerConnectionBuilder messengerConnectionBuilder;
+    private final PontonXPAdapterConfiguration config;
     private final MessengerHealth healthApi;
     private final MessengerMonitor messengerMonitor;
     private final ReentrantLock lock = new ReentrantLock();
     private final PontonRetryableStrategy retryStrategy;
+    private final PontonWebSocketConnectionState webSocketConnectionState = new PontonWebSocketConnectionState();
+    private boolean closed;
     @Nullable
     private OutboundMessageStatusUpdateHandler outboundMessageStatusUpdateHandler;
     @Nullable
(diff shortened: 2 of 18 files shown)


region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImpl.java
```diff
@@ -148,30 +171,81 @@ public PontonMessengerConnection withECMPListHandler(ECMPListHandler ecmpListHan
...
-        return healthApi.messengerStatus();
+        var messengerStatus = healthApi.messengerStatus();
+        var healthChecks = new HashMap<>(messengerStatus.healthChecks());
+        var adapterConnectionHealthCheck = webSocketConnectionState.healthCheck(
+                config.outboundConnections(),
+                config.inboundConnections(),
+                config.archiveConnections(),
+                config.connectionStatusTimeout(),
+                Instant.now()
+        );
```
fweingartshofer: issue: The config object is de-structured and each value passed individually.

suggestion: In this case it might be better to pass the whole config object instead of passing the specific counts.
This hides the implementation details of the webSocketConnectionState and allows extending the method to consider more information too in the future if needed.
Additionally, it increases readability, since having three arguments of the same type can lead to incorrect method calls.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImpl.java
```diff
@@ -148,30 +171,81 @@ public PontonMessengerConnection withECMPListHandler(ECMPListHandler ecmpListHan
...
+        return new MessengerStatus(healthChecks, messengerStatus.ok() && adapterConnectionHealthCheck.ok());
+    }
+
+    @Override
+    public void reconnectIfConnectionStale() throws TransmissionException, ConnectionException {
+        if (!webSocketConnectionState.needsReconnect(
+                config.outboundConnections(),
+                config.inboundConnections(),
+                config.archiveConnections(),
+                config.connectionStatusTimeout(),
```
fweingartshofer: todo: Same as above.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImpl.java
```diff
@@ -148,30 +171,81 @@ public PontonMessengerConnection withECMPListHandler(ECMPListHandler ecmpListHan
...
+        )) {
+            return;
+        }
+        if (!lock.tryLock()) {
+            if (LOGGER.isWarnEnabled()) {
+                LOGGER.warn(
+                        "Skipping Ponton XP adapter WebSocket reconnect because another reconnect is already running. State: {}",
+                        webSocketConnectionState.describe()
+                );
+            }
```
fweingartshofer: issue: The nesting of if-statements to log something reduces readability.

suggestion: This can be simplified by using the fluent API, this reduces nesting and increases readability.


```suggestion
        LOGGER.atWarn()
                      .addArgument(webSocketConnectionState::describe)
                      .log("Skipping Ponton XP adapter WebSocket reconnect because another reconnect is already running. State: {}");
```

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImpl.java
fweingartshofer, line 1: issue: Timestamp calls such as `Instant.now()` and `ZonedDateTime.now()` should always be used in a combination with a time zone, zone offset, or a specific clock.
Depending on the API different default clocks will be used silently.

suggestion: Pass Clock.systemUTC() or a timezone instead.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImpl.java
```diff
@@ -148,30 +171,81 @@ public PontonMessengerConnection withECMPListHandler(ECMPListHandler ecmpListHan
...
+            return;
+        }
+        try {
+            if (closed) {
+                return;
+            }
```
fweingartshofer: todo: This should be on top of the method, since it is the cheapest and most important check.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImpl.java
```diff
@@ -148,30 +171,81 @@ public PontonMessengerConnection withECMPListHandler(ECMPListHandler ecmpListHan
...
+            if (!webSocketConnectionState.needsReconnect(
+                    config.outboundConnections(),
+                    config.inboundConnections(),
+                    config.archiveConnections(),
+                    config.connectionStatusTimeout(),
+                    Instant.now()
+            )) {
```
fweingartshofer: todo: This method is called twice in the same method to do the same check.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/PontonXPAdapterConfiguration.java
```diff
@@ -29,5 +37,56 @@ public record PontonXPAdapterConfiguration(
         @Name("api.endpoint") String apiEndpoint,
         @Name("folder") String workFolder,
         @Name("username") String username,
-        @Name("password") String password
-) {}
+        @Name("password") String password,
+        @Name("inbound-connections") int inboundConnections,
+        @Name("outbound-connections") int outboundConnections,
+        @Name("archive-connections") int archiveConnections,
```
fweingartshofer: discussion: The archive connections are not used in this implementation AFAIK.
So no need to configure them.

todo: Furthermore, there should be default values for the new config values, so existing deployments don't need to be updated.
See https://docs.spring.io/spring-boot/api/java/org/springframework/boot/context/properties/bind/DefaultValue.html

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/PontonXPAdapterConfiguration.java
```diff
@@ -29,5 +37,56 @@ public record PontonXPAdapterConfiguration(
...
+    private static final int DEFAULT_INBOUND_CONNECTIONS = 1;
+    private static final int DEFAULT_OUTBOUND_CONNECTIONS = 1;
+    private static final int DEFAULT_ARCHIVE_CONNECTIONS = 0;
+    private static final Duration DEFAULT_CONNECTION_WATCHDOG_INTERVAL = Duration.ofMinutes(1);
+    private static final Duration DEFAULT_CONNECTION_STATUS_TIMEOUT = Duration.ofMinutes(2);
+
+    @ConstructorBinding
+    public PontonXPAdapterConfiguration {
+        inboundConnections = inboundConnections <= 0 ? DEFAULT_INBOUND_CONNECTIONS : inboundConnections;
+        outboundConnections = outboundConnections <= 0 ? DEFAULT_OUTBOUND_CONNECTIONS : outboundConnections;
```
fweingartshofer: issue: this silently changes the configuration without informing the operator.
suggestion: Use springs assertion mechanism to validate the correctness of the configuration. See https://jakarta.ee/specifications/bean-validation/3.0/apidocs/jakarta/validation/constraints/asserttrue

The thought behind this is, that if the operator wants to configure the values and accidentally sets the incorrect value such as "0", they should be informed about their incorrect configuration.
On the other hand, if they skip the configuration it should fall back to a valid default value.

fweingartshofer (changes requested):
Thank you for your contribution!
Looks like a good addition to make the EDA region connector more resilient.

Keep in mind that at least the build should be green before we start reviewing code, since a broken build often indicates work in progress.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImpl.java
nathaniel-boisgard: Updated healthCheck and needsReconnect to receive the complete configuration object.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImpl.java
nathaniel-boisgard: Updated this call as well. The complete configuration object is now passed.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImpl.java
nathaniel-boisgard: Updated to use the fluent SLF4J API with lazy state evaluation.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImpl.java
nathaniel-boisgard: Updated to use an injected Clock. Production supplies Clock.systemUTC(), while tests can supply a fixed clock for deterministic timeout behavior.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImpl.java
nathaniel-boisgard: Moved closed to the initial short circuit check and made it volatile. A second check after acquiring the lock remains necessary because shutdown may occur concurrently.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImpl.java
nathaniel-boisgard: I kept the second check intentionally. Connection status callbacks can update the state between the initial lock free check and acquiring the lock. Rechecking after acquiring the lock prevents reconnecting a connection that recovered during that interval. I added inline comments documenting this pattern.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/PontonXPAdapterConfiguration.java
nathaniel-boisgard: Removed archive connections from configuration and fixed the library required archive count at zero. Added Spring DefaultValue defaults for the new settings.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/PontonXPAdapterConfiguration.java
nathaniel-boisgard: Added validation that fails during binding. Omitted values receive defaults, while explicitly configured zero or negative values fail validation. Counts use Jakarta Positive and durations use property level DurationMin.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImpl.java
```diff
@@ -193,48 +199,30 @@ public void sendMessage(
...
         if (!lock.tryLock()) {
-            if (LOGGER.isWarnEnabled()) {
-                LOGGER.warn(
-                        "Skipping Ponton XP adapter WebSocket reconnect because another reconnect is already running. State: {}",
-                        webSocketConnectionState.describe()
-                );
-            }
+            LOGGER.atWarn()
+                  .addArgument(webSocketConnectionState::describe)
+                  .log("Skipping Ponton XP adapter WebSocket reconnect because another reconnect is already running. State: {}");
```
fweingartshofer: discussion: The lock might be active during other operations too, such as sending a message or closing the connection, if I understand correctly. So this might be confusing, since the lock message says that it is trying to reconnect, while it might only be sending a message, etc?

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImpl.java
```diff
@@ -193,48 +199,30 @@ public void sendMessage(
...
-            if (!webSocketConnectionState.needsReconnect(
-                    config.outboundConnections(),
-                    config.inboundConnections(),
-                    config.archiveConnections(),
-                    config.connectionStatusTimeout(),
-                    Instant.now()
-            )) {
+            // Connection-status callbacks can update the state while this thread waits for the lock. Rechecking after
+            // acquiring it prevents reconnecting a connection that recovered after the initial lock-free fast-path check.
+            if (!webSocketConnectionState.needsReconnect(config, clock.instant())) {
```
fweingartshofer: todo: this can be collapsed into one if-statement, and the condition moved to its own method, to increase readability.
This would allow reusing the method in line 210 too.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImpl.java
```diff
@@ -32,24 +32,28 @@
...
     private final MessengerConnection.MessengerConnectionBuilder messengerConnectionBuilder;
     private final PontonXPAdapterConfiguration config;
+    // The clock is injected instead of calling Instant.now(Clock.systemUTC()) directly. Production still supplies the
+    // suggested UTC clock, while tests can supply a fixed clock to verify timeout behavior without relying on wall time.
+    private final Clock clock;
     private final MessengerHealth healthApi;
     private final MessengerMonitor messengerMonitor;
     private final ReentrantLock lock = new ReentrantLock();
     private final PontonRetryableStrategy retryStrategy;
     private final PontonWebSocketConnectionState webSocketConnectionState = new PontonWebSocketConnectionState();
```
fweingartshofer: issue: The webSocketConnectionState always gets a clock.now() passed when a method is called, which bloats the API surface.
The same is true for the `PontonXPAdapterConfiguration`, it does not need to be passed as a method parameter, since it cannot change during the runtime, since it is immutable (record) and final here.

suggestion: Pass the clock and config to the this class when creating it and create the timestamps in the methods of `PontonWebSocketConnectionState` instead of passing a new timestamp for each method call.
This reduces the API surface.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/PontonXPAdapterConfiguration.java
```diff
@@ -38,55 +42,16 @@ public record PontonXPAdapterConfiguration(
...
+        @DefaultValue("2m") Duration connectionStatusTimeout
 ) {
-    private static final int DEFAULT_INBOUND_CONNECTIONS = 1;
-    private static final int DEFAULT_OUTBOUND_CONNECTIONS = 1;
-    private static final int DEFAULT_ARCHIVE_CONNECTIONS = 0;
-    private static final Duration DEFAULT_CONNECTION_WATCHDOG_INTERVAL = Duration.ofMinutes(1);
-    private static final Duration DEFAULT_CONNECTION_STATUS_TIMEOUT = Duration.ofMinutes(2);
-
     @ConstructorBinding
     public PontonXPAdapterConfiguration {
```
fweingartshofer: todo: can be removed.

region-connectors/region-connector-at-eda/src/test/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImplTest.java
```diff
@@ -30,7 +34,11 @@ class PontonMessengerConnectionImplTest {
             "http://localhost:8081",
             "work",
             "user",
-            "password"
+            "password",
+            1,
+            1,
+            java.time.Duration.ofMinutes(1),
+            java.time.Duration.ofMinutes(2)
```
fweingartshofer: todo: import `Duration`

region-connectors/region-connector-at-eda/src/test/java/energy/eddie/regionconnector/at/eda/ponton/messenger/WebClientMessengerHealthTest.java
```diff
@@ -37,7 +37,11 @@ static void setUp() throws IOException {
                 "http://localhost:" + mockBackEnd.getPort(),
                 "folder",
                 "username",
-                "password"
+                "password",
+                1,
+                1,
+                java.time.Duration.ofMinutes(1),
```
fweingartshofer: todo: import `Duration`

region-connectors/region-connector-at-eda/src/test/java/energy/eddie/regionconnector/at/eda/ponton/messenger/WebClientPontonTokenProviderTest.java
```diff
@@ -52,7 +52,11 @@ static void setUp() throws IOException {
                 "http://localhost:" + mockBackEnd.getPort(),
                 "folder",
                 "user",
-                "password"
+                "password",
+                1,
+                1,
+                java.time.Duration.ofMinutes(1),
+                java.time.Duration.ofMinutes(2)
```
fweingartshofer: todo: import `Duration`

fweingartshofer (changes requested):

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImpl.java
nathaniel-boisgard: Yes that is correct. I have adapted the message accordingly, it was poorly worded.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImpl.java
nathaniel-boisgard: Extracted the condition into `shouldSkipReconnect()` and reused it in both locations. The second evaluation remains necessary because the connection state can change before the lock is acquired.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImpl.java
nathaniel-boisgard: Moved the immutable configuration and clock into `PontonWebSocketConnectionState`. Timestamp creation is now handled internally.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/PontonXPAdapterConfiguration.java
nathaniel-boisgard: Removed the empty constructor and the unnecessary `ConstructorBinding` annotation.

region-connectors/region-connector-at-eda/src/test/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImplTest.java
nathaniel-boisgard: Replaced the qualified reference with a `Duration` import.

region-connectors/region-connector-at-eda/src/test/java/energy/eddie/regionconnector/at/eda/ponton/messenger/WebClientMessengerHealthTest.java
nathaniel-boisgard: Replaced the qualified references with a `Duration` import.

region-connectors/region-connector-at-eda/src/test/java/energy/eddie/regionconnector/at/eda/ponton/messenger/WebClientPontonTokenProviderTest.java
nathaniel-boisgard: Replaced the qualified references with a `Duration` import.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonWebSocketConnectionState.java
```diff
@@ -0,0 +1,107 @@
...
+            int outboundConnectionCount,
+            int inboundConnectionCount
+    ) {
+        this.outboundConnectionCount = outboundConnectionCount;
+        this.inboundConnectionCount = inboundConnectionCount;
+        lastConnectionStatusChangedAt = clock.instant();
+    }
+
+    synchronized boolean needsReconnect() {
+        if (!receptionStarted) {
```
fweingartshofer: todo: Can be replaced by a null check on lastStartAt, since they both are first set at the same time.
That way receptionStarted can be removed from the class.
```suggestion
        if (lastStartAt == null) {
```

region-connectors/region-connector-at-eda/src/test/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImplTest.java
fweingartshofer: Did you forget to commit it? 😄

fweingartshofer:
todo: There are some smaller issues from sonar: 
image

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonWebSocketConnectionState.java
nathaniel-boisgard: Removed `receptionStarted`. `lastReceptionStartAt` is now the authoritative state.

region-connectors/region-connector-at-eda/src/test/java/energy/eddie/regionconnector/at/eda/ponton/messenger/PontonMessengerConnectionImplTest.java
nathaniel-boisgard: The qualified references were still present. Corrected.

nathaniel-boisgard:
Extracted the three nested mocks into local variables.

fweingartshofer (approved):
Nice work!

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/ponton/PontonXPAdapterConfiguration.java
```diff
@@ -29,5 +40,14 @@ public record PontonXPAdapterConfiguration(
...
-) {}
+        @Name("password") String password,
+        @Name("inbound-connections") @Positive @DefaultValue("1") int inboundConnections,
+        @Name("outbound-connections") @Positive @DefaultValue("1") int outboundConnections,
+        @Name("connection-watchdog.interval")
+        @DurationMin(nanos = 1)
+        @DefaultValue("1m") Duration connectionWatchdogInterval,
+        @Name("connection-watchdog.status-timeout")
+        @DurationMin(nanos = 1)
+        @DefaultValue("2m") Duration connectionStatusTimeout
```
re1, line 51: While mostly internal, the new properties should probably be documented in the framework documentation under `docs/1-running/region-connectors/region-connector-at-eda.md` published [here](https://architecture.eddie.energy/framework/1-running/region-connectors/region-connector-at-eda.html).

fweingartshofer: Good catch!

fweingartshofer (changes requested):

fweingartshofer (approved):

fweingartshofer (changes requested):
The commit messages are not fitting the conventional commit spec, update the violating message, and we are good to go.

fweingartshofer (approved):

re1 (approved):

PR #2584, opened 2026-08-17
feat(aiida): Only data sources with a matching meter ID are displayed for the FCA outbound permission.

### LINKED ISSUE #2547: AIIDA: Limit data source select options when accepting an FCA outbound permission by meter ID
## Short Description
**As** an AIIDA user **I want** to only have data sources as select options which meter ID matches the FCA **so that** I can more easily find the correct data source to pick for my FCA.

## Acceptance Criteria
- [x] Selectable data sources match the permission's meter ID if one is provided
- [x] If only one data source matches the meter ID, select it by default

stefanpenzinger:
This PR adds an optional request parameter to filter outbound data sources by their physical meter ID.
This is relevant in the context of FCAs in order to display the user the relevant data sources matching the physical meter ID.
If there is only one matching result, the data source is pre-selected.


re1 (approved):
Looks good 👍

Issue #2583, opened 2026-08-14
Elering region connector: Create module and related classes for the region connector


AyaKhaled1:
## Short Description
Create a new submodule for the region connector and add some basic dependencies.
Furthermore, add all classes required so the region connector is picked up by the core module.

## Acceptance Criteria
- [ ] `region-connector-ee-elering` module is created
- [ ] Region connector is picked up by the core
- [ ] Region connector is documented

sgrue:
Blocked until e-residency issue is clarified.

PR #2585, opened 2026-08-17
GH-2583 Elering region connector: create and document the region-conn…

AyaKhaled1:
…ector-ee-elering module


fweingartshofer (commented):
The entry for the permission-administrator.json and metered-data-administrator is missing.

Issue #13, opened 2023-03-06
Create a github action to test and build the gradle project


fweingartshofer:
Since we want to use CI/CD as soon as possible we need actions to run tests and build the project.

PR #2586, opened 2026-08-17
Eddie Example App GH-13: Simplify first contact with Example App and local EDDIE

Flo-Auzi:
- Add example app user to EDDIE keycloak realm
- Change AIIDA specific ports and network in order to be able to locally run in combination with the example app


env/keycloak/EXAMPLE_APP-realm.json
```diff
@@ -0,0 +1,46 @@
+{
+  "realm": "EXAMPLE_APP",
```
re1, line 2: The realm name being in caps is not a convention. They were just created that way because EDDIE and AIIDA are both acronyms. You can just name the file `example-app-realm.json` and the realm "Example App".

aiida/docker/compose.yml
```diff
@@ -113,4 +116,9 @@ volumes:
   db_data:
   emqx_data:
   keycloak_data:
-  aiida_security:
\ No newline at end of file
+  aiida_security:
+
+networks:
```
re1, line 121: What is the extra network for?

re1 (commented):
Now seeing the example app configuration here I think we should wait with this integration until we reconsidered some of its scope. Particularly that it needs MQTT configuration where I feel it should not.

The example app should be able to reuse the existing Postgres database. Take a look at `env/postgres/create-emqx-user.sh`.

Flo-Auzi:
> Now seeing the example app configuration here I think we should wait with this integration until we reconsidered some of its scope. Particularly that it needs MQTT configuration where I feel it should not.
> 
> The example app should be able to reuse the existing Postgres database. Take a look at `env/postgres/create-emqx-user.sh`.

I was told that when this example app was initially developed, there had not been a possibility to get real time data via kafka - so that's why MQTT was integrated that way.

PR #2595, opened 2026-08-24
feat(aiida): Display when the latest message was sent/received for inbound/outbound permissions

### LINKED ISSUE #2163: AIIDA: Show timestamp of latest message sent/received for permissions
## Short Description
**As** an AIIDA user **I want** to know for inbound permission when I received, and for outbound permission when I sent the latest message, **so that** I can keep track of when things happen.
This information should be made available in the permission details.

## Acceptance Criteria

- [ ] Show a timestamp for permissions
- [ ] Adapt the text (received/sent) for inbound/outbound permissions

stefanpenzinger:
It adds a 'last message received'/'last message sent' field to the permission details for 'inbound'/'outbound' permissions.
When the details of a permission are viewed, the front end periodically checks whether a new message has been sent or received.
For this, the frontend calls the backend to derive the next expected message. If no message is expected (e.g., transmission disabled), the frontend uses a default value instead. 
If the permission has reached a terminal status, the periodic check will not be executed again.


aiida/src/main/java/energy/eddie/aiida/dtos/record/NextExpectedTransmissionDto.java
```diff
@@ -0,0 +1,14 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package energy.eddie.aiida.dtos.record;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import jakarta.annotation.Nullable;
+
+import java.time.Instant;
+
+public record NextExpectedTransmissionDto(
```
re1: Probably personal preference, but I usually avoid defining DTOs with a single field and suggest using it as a scalar resource instead. I assume your reason to do so is to have it be a named JSON field for extensibility. However, the effort needed to change it later is so small that I do not see the benefit in adding the abstraction layer now.

aiida/ui/src/constants/data-need-type.ts
re1, line 1: I think the data need types should not be stored in a constant but as a TypeScript type. If you are not iterating over its values I see no benefit in having it externalized. Unlike Java (for reference), as long as you have type information, a statement like `if (permission.dataNeed.type === 'inbound') {` is closer to an enum comparison than a string comparison. `ts-lint` will fail if your right-hand side argument is not a valid instance of the data need type. You are already sure that if the type is refactored the comparison has to be refactored as well. IDEs also have excellent TypeScript support so they will recommend you possible values and highlight incorrect ones as errors. Could be that I fail to see your motivation though.

aiida/ui/src/composables/last-message-refresh.ts
```diff
@@ -0,0 +1,103 @@
...
+) {
+  const lastMessageAt = ref<Date | null>(null)
+  let refreshTimer: ReturnType<typeof setTimeout> | undefined
+
+  const isTerminalStatus = computed(() => {
+    const status = STATUS[getPermission().status]
+    return !status?.isActive && !status?.isOpen
+  })
+
+  const fetchLastMessageTimestamp = async () => {
```
re1: Refetching the full latest message just to grab its timestamp and not even having it in the right format seems like overkill on the client side.

If I interpret it correctly, you only use the information when the next message is planned to arrive to inform when to refetch, right? Some ideas:
1 Why not just pass the effective transmission schedule to the frontend and use that information to inform the refetch delay?
2. Maybe do a separate endpoint to fetch the last updated timestamp instead of fetching the full message every time.
3. If we actually want this to update in real time, why not invert the flow and let the UI subscribe to an SSE endpoint that pushes the timestamp of the latest message once it arrives?

aiida/src/main/java/energy/eddie/aiida/dtos/record/NextExpectedTransmissionDto.java
stefanpenzinger: I did it because it is consistent with our other endpoints.
Can change it to scalar if you insist :slightly_smiling_face:

fweingartshofer: I like that approach, since the timestamp can be null, I don't think you could really represent with just the scalar. I guess 404 would be good, if that makes semantically sense.

re1 (commented):
Not very happy with how the `api.ts` has grown over time or the complexity needed for the `last-message-refresh` composable. However, resolving these issues requires some significant refactoring outside the scope of this feature.

Not sure if I interpret this correctly, but it seems the database breaks down with a larger number of permissions as I get 404 on request and 500 in the AIIDA logs with an SQL exception "Connection Not Available".

Tested with 137 permissions (only 2 active) and about 230 records. Standard compose setup as defined in the repo with local EDDIE and AIIDA. I attached my database for you so you do not have to create the permissions yourself.

Ideally, the reactive endpoint would only have to interact with the database to load existing permissions and their latest record once. With indexing this should be a fairly easy task.

[AIIDA-2026_08_28_13_00_32-dump.sql](https://github.com/user-attachments/files/31551624/AIIDA-2026_08_28_13_00_32-dump.sql)

re1 (changes requested):

re1 (approved):
Good refactoring on the recent update. `spring.jpa.open-in-view=false` was overdue. We might want to watch for potential errors from this change, just to be sure.

Local testing worked for me.

Remaining E2E failure is unrelated.

PR #2598, opened 2026-08-25
eda region connector: fix econ request and update demo page to filter data needs by name

fweingartshofer:
Changes to the demo page:
image

### DIFF
--- a/core/src/main/resources/templates/demo.html
+++ b/core/src/main/resources/templates/demo.html
@@ -50,9 +50,24 @@
       <a href="https://architecture.eddie.energy/framework/1-running/eddie-button/eddie-button.html">Operation Manual</a>.
     </p>
     <form>
+
+      <label>
+        <span>Data Need Filter</span>
+        <input
+          aria-controls="data-need-select"
+          aria-label="Data need filter"
+          placeholder="Filter..."
+          type="search"
+        />
+      </label>
       <label>
         <span>Data need (required)</span>
-        <select autocomplete="off" multiple name="data-need-id"></select>
+        <select
+          autocomplete="off"
+          id="data-need-select"
+          multiple
+          name="data-need-id"
+        ></select>
         <small>
           Multiple data needs can be selected, but not all region connectors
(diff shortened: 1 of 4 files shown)


core/src/main/resources/templates/demo.html
```diff
@@ -52,6 +52,7 @@
     <form>
       <label>
         <span>Data need (required)</span>
+        <input placeholder="Filter…" type="search" />
```
re1: Not particularly relevant for a demo page, but I want to note that this is not valid HTML since this label around it now has two inputs associated with it. Every input can only be associated with one label.

A valid variant would be:

```html
<label>
    <span>Data need filter</span>
    <input placeholder="Filter…" type="search" aria-controls="data-need-select" />
</label>
<label>
  <span>Data need (required)</span>
  <select id="data-need-select" autocomplete="off" multiple name="data-need-id"></select>
  <small>
    Multiple data needs can be selected, but not all region connectors
    support multiple data needs.
  </small>
</label>
```

You could also add the `aria-label` attribute to the input directly if you find the span text unnecessary (which I do).

```html
<input placeholder="Filter…" type="search" aria-label="Data need filter" aria-controls="data-need-select" />
<label>
  <span>Data need (required)</span>
  <select id="data-need-select" autocomplete="off" multiple name="data-need-id"></select>
  <small>
    Multiple data needs can be selected, but not all region connectors
    support multiple data needs.
  </small>
</label>
```

Also, please add a screenshot for UI changes, so reviewers do not have to check out the code for smallish changes like this.

re1 (commented):
Unrelatedness and lack of mention in PR name makes me think the demo button commit was accidental?

core/src/main/resources/templates/demo.html
fweingartshofer: WIll add screenshots, I forgot to change the default title for the PR sorry. :(
It just got annoying scrolling through the different data needs.

re1 (approved):

PR #2602, opened 2026-09-01
feat(at-eda): prefix conversation IDs for Ponton routing

nathaniel-boisgard:
## What changed

- add the optional `region-connector.at.eda.conversation-id.prefix` configuration property
- prefix generated EDA grouping IDs used as conversation and message IDs for CM requests and CP retransmissions as this was the way it worked before
- generate prefixed message IDs for CM revokes while preserving their original conversation IDs
- validate prefixes against the EDA character set and 35-character grouping-ID limit at startup
- add configuration, serialization, routing, and boundary tests
- document the new configuration property

## Why

A Ponton X/P Messenger instance can have only one default adapter. Previously,
generated conversation IDs did not identify the EDDIE instance or adapter that
initiated the conversation.

This meant that multiple EDDIE instances could not reliably use the same
Ponton X/P Messenger instance. The messenger could not distinguish
conversations belonging to different adapters and therefore could not route
incoming messages to the adapter that initiated a specific conversation.

With a unique prefix configured for each EDDIE instance, the conversation ID
provides Ponton with a stable identifier for associating conversations with
their originating adapters. Ponton can then route subsequent incoming messages
for each conversation to the correct adapter.

In the case of a Revocation, we keep the old conversation ID, as it is undocumented if this might have a negative effect (see definition of process https://www.ebutilities.at/prozesse/651), if we have the situation where a prefix was introduced after a request process was started, resulting in un-prefixed conversations being referenced, but a prefixed conversation ID given.

## Impact

Multiple EDDIE instances can share one Ponton X/P Messenger instance when each
instance uses a unique prefix, for example:

`region-connector.at.eda.conversation-id.prefix=EDDIE1`

The default prefix is empty, so existing installations retain their current
behavior without configuration changes.

Prefixing applies to newly created conversations. Existing persisted
conversations retain their original conversation IDs because changing them
would break correlation with EDA.

## Validation

- full AT EDA connector check
- configuration binding and startup validation tests
- eligible-party and energy-community conversation-ID tests
- CM request, CP retransmission, and CM revoke serialization tests
- maximum grouping-ID length tests
- JaCoCo coverage verification


region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/config/AtConfiguration.java
```diff
@@ -4,24 +4,41 @@
...
+        @Pattern(regexp = CONVERSATION_ID_PREFIX_PATTERN,
+                message = "must contain only ASCII letters and digits")
+        @DefaultValue("") String conversationIdPrefix
+) {
+    public static final String CONVERSATION_ID_PREFIX_PATTERN = "[0-9A-Za-z]*";
+
     @AssertTrue(message = "If the energy community ID has been set, the energy community party ID must also be set")
     public boolean isEnergyCommunityIdSetMustBeEnergyCommunityPartyIdAlsoSet() {
-        return energyCommunityId != null && energyCommunityPartyId != null;
+        return energyCommunityId == null || (energyCommunityPartyId != null && !energyCommunityPartyId.isBlank());
```
fweingartshofer: todo: If the energyCommunityPartyIs is checked for blanks, the same should happen to the energyCommunityId

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/handlers/TerminationHandler.java
```diff
@@ -47,10 +57,17 @@ public void accept(PermissionEvent permissionEvent) {
...
         }
         AtPermissionRequest permissionRequest = request.get();
-        var revoke = new CCMORevoke(permissionRequest,
-                                    atConfiguration.eligiblePartyId(),
-                                    "Terminated by the Eligible Party");
         try {
+            var messageId = groupingIdFactory.create(
+                    AtConfiguration.PartyIdType.ELIGIBLE_PARTY,
+                    ZonedDateTime.now(EdaRegionConnectorMetadata.AT_ZONE_ID)
+            );
```
fweingartshofer, line 64: discussion: #2599 changes the revocation process, we should coordinate these changes, so nothing breaks or leads to unintended behaviour.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/permission/request/events/ValidatedEventFactory.java
```diff
@@ -48,16 +49,16 @@ public ValidatedEvent createValidatedEvent(
...
+        var cmRequestId = new CMRequestId(groupingId).toString();
 
         return new ValidatedEvent(
                 permissionId,
                 start,
                 end,
                 granularity,
                 cmRequestId,
-                messageId,
+                groupingId,
```
fweingartshofer: question: What is the intend behind persisting the grouping ID and not the message ID?
todo: The database column is called conversationId, which will lead to missunderstandings.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/permission/request/events/ValidatedEventFactory.java
```diff
@@ -48,16 +49,16 @@ public ValidatedEvent createValidatedEvent(
         var type = dataNeedCalculation instanceof CESUJoinRequestDataNeedResult
                 ? AtConfiguration.PartyIdType.ENERGY_COMMUNITY
                 : AtConfiguration.PartyIdType.ELIGIBLE_PARTY;
-        var messageId = new MessageId(configuration.partyIdFor(type), created).toString();
-        var cmRequestId = new CMRequestId(messageId).toString();
+        var groupingId = groupingIdFactory.create(type, created);
+        var cmRequestId = new CMRequestId(groupingId).toString();
```
fweingartshofer: todo: This variable is never used.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/requests/EdaGroupingIdFactory.java
```diff
@@ -0,0 +1,62 @@
...
+    private static final Pattern VALID_PREFIX = Pattern.compile(AtConfiguration.CONVERSATION_ID_PREFIX_PATTERN);
+
+    private final AtConfiguration configuration;
+
+    public EdaGroupingIdFactory(AtConfiguration configuration) {
+        this.configuration = requireNonNull(configuration);
+        var prefix = requireNonNull(configuration.conversationIdPrefix(), PREFIX_PROPERTY + " must not be null");
+        if (!VALID_PREFIX.matcher(prefix).matches()) {
+            throw new IllegalArgumentException(PREFIX_PROPERTY + " must contain only ASCII letters and digits");
+        }
```
fweingartshofer: todo: These checks are already present in the AtConfiguration, can be removed, so they do not become out of sync.

fweingartshofer (changes requested):
The broken build has been fixed on main, you can rebase to fix it here too.

On another note, did you test this for requests, where data will be available in the future? For example, I request the data for the coming 3 months and once it is available the DSO sends it to the EDDIE instance.
Previously, we had the problem that conversationIDs are not reused in these cases, making the routing obsolete.

nathaniel-boisgard:
> On another note, did you test this for requests, where data will be available in the future? For example, I request the data for the coming 3 months and once it is available the DSO sends it to the EDDIE instance. Previously, we had the problem that conversationIDs are not reused in these cases, making the routing obsolete.

To test for future outside eventualities is pretty hard :) 
However - the change running oin a deployed fork is now running for a month with a stable conversation ID issued in the permission process.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/requests/EdaGroupingIdFactory.java
nathaniel-boisgard: Agreed. The duplicate null and character-set checks were removed. The combined ID-length validation remains because it depends on both the prefix and the selected party ID.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/permission/request/events/ValidatedEventFactory.java
nathaniel-boisgard: cmRequestId is passed to ValidatedEvent, persisted, included in the outbound request, and used for response correlation. Could you clarify if you meant a different variable?

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/handlers/TerminationHandler.java
nathaniel-boisgard: Agreed. When integrating #2599, the selected party type must be used consistently for both the revoke sender and the generated message ID.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/config/AtConfiguration.java
nathaniel-boisgard: Agreed. Validation and a corresponding startup test have been added.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/permission/request/events/ValidatedEventFactory.java
nathaniel-boisgard: Historically, this generated value has been used as both MessageId and ConversationId. I inferred that it groups messages belonging to the same process, hence groupingId. To avoid confusion, I renamed the local variable back to messageId.

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/permission/request/events/ValidatedEventFactory.java
nathaniel-boisgard: can you clarify @fweingartshofer ?

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/handlers/TerminationHandler.java
fweingartshofer: I think this PR is ready to go, so I will rebase once this is merged.

fweingartshofer (approved):

PR #2604, opened 2026-09-02
GH-2597 eda region connector import pre existing permissions

### LINKED ISSUE #2597: EDA Region Connector: Import pre-existing permissions
## Short Description

As a eligible party I might already have consent to access validated historical data or an active participant of an energy community.
The push-based approach in Austria makes the region connector eligible to import pre-existing permissions.
This would allow the EDA region connector to recognize data incoming from the ponton messenger and map it to the imported permission.

## Acceptance Criteria
- [x] There is a permission import endpoint that allows eligible parties to import existing permissions.
- [x] Only certain data needs are allowed, such as validated historical data  or CESU data needs.
- [x] The EP can specify required items, such as initial creation of the permission, metering point ID, and the Consent ID

fweingartshofer:


docs/1-running/region-connectors/region-connector-at-eda.md
```diff
@@ -29,16 +29,14 @@ depends on the way you deploy the region connector.
...
+| `region-connector.at.eda.energy-community-id`                           | The energy community id that is used to identify the energy community when requesting to add members to the energy community. See the code snippet below for examples.                                                                                                                                                                                                                                                                                                              |
+| `region-connector.at.eda.energy-community-party-id`                     | The energy community party id that is used to identify the energy community in the Austrian energy market system. It is used as the sender in the Ponton X/P Messenger                                                                                                                                                                                                                                                                                                              |
```
re1, line 39:
```suggestion
| `region-connector.at.eda.energy-community-party-id`                     | The energy community party id that is used to identify the energy community in the Austrian energy market system. It is used as the sender in the Ponton X/P Messenger.                                                                                                                                                                                                                                                                                                              |
```

docs/1-running/region-connectors/region-connector-at-eda.md
```diff
@@ -29,16 +29,14 @@ depends on the way you deploy the region connector.
...
+| `region-connector.at.eda.ponton.messenger.api.endpoint`                 | Endpoint of the XP Messenger REST API. Default should be `<hostname>[:<api-port>]/api`                                                                                                                                                                                                                                                                                                                                                                                              |
```
re1, line 32:
```suggestion
| `region-connector.at.eda.ponton.messenger.api.endpoint`                 | Endpoint of the XP Messenger REST API. Default should be `<hostname>[:<api-port>]/api`.                                                                                                                                                                                                                                                                                                                                                                                              |
```

region-connectors/region-connector-at-eda/src/main/java/energy/eddie/regionconnector/at/eda/dto/validation/MeterReadingPairValidator.java
re1, line 1: You could remove the need for this particular validator by grouping the start and end fields into a record for the DTO. Inside your `PermissionRequestToImport` you would have something like this:

```java
@Nullable
MeterReadingTimeframe meterReadingTimeframe;

private record MeterReadingTimeframe(ZonedDateTime start, ZonedDateTime end) {}
```

The event could stay the same if preferred.

Personally, if you do not plan on reusing the validations, I would just place the checks in the controller and throw a custom or validation exception, removing 6 classes and 3 tests. Downside is that the validation logic is split between DTO and controller.

fweingartshofer: I like the convenience of the additional validators, since it makes the controller very easy to read and I can reuse the controller advisors.
So I will stick to the dedicated validators, to keep the controller slim.

PR #2606, opened 2026-09-02
GH-2588 Allow the EP to set default connection limits for AIIDA permissions

### LINKED ISSUE #2588: AIIDA: Allow EP to set default fallback connection limits per permission
## Short Description

**As** an eligible party **I can** set default connection limits that my customer should fall back on in case they do not receive my connection limit market documents, **so** limits are not violated in case there is an issue with the data exchange.

To achieve this, the EP should be able to set an initial default connection limit in the permission request. The default limit cannot be set in the data need as limits are specific to the customer/meter. The EP should also be able to update the default limit via a [Permission Command](https://architecture.eddie.energy/aiida/1-running/data-need.html#permission-commands).

The default limits are optional. If no default limit exists and no limit is found for a requested time frame, no limit is to be returned via the API.

While particularly relevant for flexible connection agreements, this feature is also intended to be used without the FCA context. For example, with controllable units.

This feature is separate from, and we allow the EP to do that, "de facto" default limits by sending a connection limit document (min-max envelope) with a limit spanning the entire duration of the permission.

## Acceptance Criteria
- [ ] Permission requests may include a default connection limit
- [ ] A permission command exists to update the default connection limit
- [x] Connection Limit API returns the default limit if no limit was saved for a specific time frame
- [x] Connection Limit API and related functionality highlights when the default is used instead of a document value
- [x] The use of the default limit can be traced in code and API for debugging purposes

re1:
Implements #2588.

Finally adds a UI test for connection limits :tada:

## Motivation

Default limits are set _for a permission_ on creation or via _permission commands_. Actual limits are set post creation via (CIM) _market documents_. The goal is to provide a fallback on permission level in case (1) the permission was just created, and no messages were sent or received yet, or (2) there is a communication issue on market document level. These goals separate this feature from maintaining a "de facto" default limit by sending market documents with a limit spanning the entire duration of the permission.

## Assumptions

- Default limits can be passed in the _permission request_ or set via _permission commands_.
	- Passing default limits as EDDIE button attributes is possible, but they could be modified or misinterpreted in the browser.
- Default limits can be set on any permission but are only used to calculate effective connection limits if the permission
	- is active (accepted, waiting, streaming),
	- is inbound and accepts the right schema.

## Questions

> **Prevent** setting default limits for permissions that do not receive min-max envelopes?
> - **Yes**: AIIDA ensures on ingestion that only limit permissions can have default limits. Customers cannot be shown limits for non-limit permissions. Querying can be less defensive. Usage is limited but intentional.
> - **No**: If we trust the EP to use limits on non-limit permissions intentionally, we can remove validation logic and allow some creative uses:
>   - Permissions defining limits for a CU that are only updated via permission commands instead of market documents.

Yes, we want to limit that feature to inbound limit permissions and specifically target other use cases separately.

 > Allow semantically invalid default limits (min >= max)?
>   - Responsibility of AIIDA, the EP, or the customer?

No, we want to prevent invalid limits in general and inform the EP on request or through acknowledgements that they made a mistake.

## Out of scope

Did not bother to clean up defaults outside the permission time frame or clamping limits to the permission time frame in general yet. I would implement this as part of the calculation to avoid handling wiring of accept/revoke/terminate events and edge-cases. However, this requires some restructuring that is only vaguely related to the default limits.

## Screenshots

image

demo


aiida/src/main/java/energy/eddie/aiida/dtos/PermissionDetailsDto.java
```diff
@@ -77,6 +98,17 @@ private void unpackPermissionRequest(Map<String, String> permissionRequest) {
...
+        }
+
+        var maxLimitKwString = permissionRequest.get("max_limit_kw");
+        if (maxLimitKwString != null) {
+            maxLimitKw = BigDecimal.valueOf(Double.parseDouble(maxLimitKwString));
+        }
+
         start = LocalDate.parse(permissionRequest.get("start"));
         end = LocalDate.parse(permissionRequest.get("end"));
     }
```
fweingartshofer, line 114: question: Why is there not a DTO that matches the fields extracted by the map?

aiida/src/main/java/energy/eddie/aiida/services/PermissionService.java
```diff
@@ -474,14 +483,29 @@ private Permission updatePermissionWithDetails(
...
+        if (permission.minLimitKw() != null || permission.maxLimitKw() != null) {
+            var defaultLimit = new ConnectionLimitDefault(permission,
+                                                          permission.meterId(),
+                                                          Objects.requireNonNull(permission.startTime()),
+                                                          null,
+                                                          permission.minLimitKw(),
+                                                          permission.maxLimitKw());
+            connectionLimitDefaultRepository.save(defaultLimit);
+        }
+    }
```
fweingartshofer, line 508: question: The if verifies that at least one of the min-max limits are not null, but will update both limits, even when one of time might be null. Is this intended?

aiida/src/main/resources/db/aiida/migration/V9_8__add_default_limits_to_permission.sql
```diff
@@ -0,0 +1,19 @@
...
+(
+    id            BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+    permission_id uuid        NOT NULL REFERENCES permission (permission_id),
+    meter_id      TEXT        NOT NULL DEFAULT '',
+    start         timestamptz NOT NULL,
+    replaced_at   timestamptz NULL,
+    min_limit_kw  DECIMAL     NULL,
+    max_limit_kw  DECIMAL     NULL,
+    CONSTRAINT connection_limit_default_valid_interval CHECK (replaced_at IS NULL OR start < replaced_at)
+);
```
fweingartshofer, line 17: issue: This will delete all data already present, a non-destructive migration would retain the data.

aiida/src/main/java/energy/eddie/aiida/dtos/PermissionDetailsDto.java
re1: Most probably but I assume there is a reason we have this. I am currently collecting items for a refactoring session in October and would postpone inspection until then.

aiida/src/main/java/energy/eddie/aiida/services/PermissionService.java
re1: Yes. The semantics are a bit tricky because null is a valid option that means "no limit in that direction". If we were to ignore null then the EP would have no way of _deleting_ a previous limit, therefore we disallow updating one limit without updating the other.

I did consider the alternative of requiring both or neither to be set, but it was stated in a refinement that we want to allow either on their own → "Stay above 3kW load, but you can go however high you want".

aiida/src/main/resources/db/aiida/migration/V9_8__add_default_limits_to_permission.sql
re1: Should have commented on it. There previously was no way to put anything into that table, so there is no way for a deployment to have any rows in there. It was added early to use it in the effective limit calculations added with #2548.

aiida/src/main/java/energy/eddie/aiida/services/PermissionService.java
fweingartshofer: So if I understand that correclty:
„If one or both are not null, set them both as new connection limit, even when it means deleting one, but I cannot delete both“.

re1: For the permission command route, the EP _can_ delete a connection limit by sending both null. The `persistLimitDefaults` you referenced is just for _creation_ so it will be called exactly once per permission when the permission is created. Therefore, there is nothing to update.

Initial Limits → Permission ← Permission Command Limits

fweingartshofer: Got it, thanks for the explanation.

stefanpenzinger (commented):
Great job, nothing to add 👍

re1:
As our E2E test runner is currently not working I am attaching a screencast of the successful E2E tests here:
https://github.com/user-attachments/assets/15c63de8-a954-4230-960c-b4e9e6568c47

stefanpenzinger (approved):
Great job 👍 😄

